Issue #2, opened 2021-05-19
Followup discussion from passport-saml PR


kriss1897:
PR: https://github.com/node-saml/passport-saml/pull/574

cjbarth:
The [node-saml](https://github.com/node-saml/node-saml) and [passport-saml](https://github.com/node-saml/passport-saml) projects are accepting all PRs that actually complete the separateion of the Node-SAML stuff from the Passport stuff. We wish to make a pair of 4.0.0 releases that complete this separation and where [node-saml](https://github.com/node-saml/node-saml) is a dependency of [passport-saml](https://github.com/node-saml/passport-saml).

kriss1897:
@zoellner @cjbarth Guys, who will be able to review the PRs? https://github.com/node-saml/passport-saml/pull/593 and https://github.com/node-saml/node-saml/pull/5

PR #3, opened 2021-05-19
remove passport-saml code and tests

zoellner:
# Description

move node-saml code and tests up one level to src/ and test/
create passport-saml-types
reset changelog
rename package to node-saml

# Checklist:

- Issue Addressed: #2
- Link to SAML spec: NA
- Tests included? Yes
- Documentation updated? No, separate PR is already started


CHANGELOG.md
```diff
@@ -1,458 +1,11 @@
 # Changelog
 
-## v2.2.0 (2021-04-23)
+## v3.0.0 (2021-05-14)
```
cjbarth, line 3: Do we want to keep the history for those that want to see where this came from? Especially since we auto-gen the changelog (though that can be configured)?

src/passport-saml-types.ts
```diff
@@ -0,0 +1,10 @@
+import * as passport from "passport";
```
cjbarth, line 1: I'm not going to push too hard on such an early PR, but I really don't think this project should have anything `passport`-related in it. What is this for?

cjbarth (changes requested):

src/passport-saml-types.ts
zoellner: turned out that there was still one cross dependency in the type definitions. this is handling that for now

CHANGELOG.md
zoellner: depends a bit on where we are going to publish this. npm package `node-saml` is currently taken by my previous fork/strip down of passport-saml but I'm happy to release that name. As that one is already at 3.0 we'd have to go to 4.0 for the first release from this repo. I thought that the old changelog is not needed here and have therefore referenced the passport-saml changelog for anything prior to what is going to be released from this repo.

Or we have to use al alternate npm package name, e.g. `@node-saml/node-saml`

cjbarth: I was expecting that we'd release a pair of 4.0.0 when we do the separation and then we could publish over the `node-saml` package name like you suggest.

src/passport-saml-types.ts
cjbarth: Sure thing... that is why we should probably release a pair of 4.0. I'm not against this being a rather quick release if we can separate out everything. We had good success with keeping the 2.x branch alive while we sorted out the 3.0 stuff. We can do the same with keeping the 3.x branch alive while we build toward a 4.0 release.

zoellner: agreed. and someone who is deeper into the passport world and typescript should probably look at these type definitions.

kriss1897: If we are using `passport` here then why is `passport` removed  from package dependencies.

zoellner: it should be removed from dependencies. it's only here in the types because those types are still somewhat mingled. if someone who's into typescript could clean them up that'd be great. I'm more a JS guy

cjbarth: Let's get this landed and then we'll have a clearer picture of what needs to be done to clean this up.

package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "passport-saml",
+  "name": "node-saml",
   "version": "3.0.0",
-  "description": "SAML 2.0 authentication strategy for Passport",
+  "description": "SAML 2.0 authentication strategy for Node.js",
```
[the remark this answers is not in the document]
cjbarth: I agree with this. Why not make this small change and then we'll land this and prepare for pass 2.

CHANGELOG.md
cjbarth: I found a problem with [gren](https://github.com/github-tools/github-release-notes) which is affecting this repo. We'll have to revisit change logs when [this](https://github.com/github-tools/github-release-notes/issues/116) gets fixed.

cjbarth (approved):

PR #5, opened 2021-05-21
Add assertion attributes to child object on profile (passport-saml#543)

kriss1897:
This attributes are also mounted to profile directly in a non conflicting way.

Duplicate PR of: https://github.com/node-saml/passport-saml/pull/593

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -24,7 +24,11 @@ jobs:
       - run: npm ci
       - run: npm test
       - run: npm ci
-      - run: npm --depth 9999 update
+      - run: npm update
+      # Remove the node_modules and re-install with updated package-lock.json
+      # because npm update command breaks some of the npm-cli packages
+      - run: rm -rf node_modules
+      - run: npm ci
       - run: npm test
         env:
           CI: true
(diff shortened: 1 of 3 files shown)


kriss1897:
This failed build on 16.x is also failing on master. Will check and raise a PR to fix

.github/workflows/workflow.yml
```diff
@@ -24,7 +24,11 @@ jobs:
       - run: npm ci
       - run: npm test
       - run: npm ci
-      - run: npm --depth 9999 update
+      - run: npm update
```
cjbarth, line 27: How is this then working in `passport-saml` if it isn't working here in `node-saml`?

cjbarth:
@kriss1897 Do you have a link to the bug in `npm`?

.github/workflows/workflow.yml
kriss1897: I tried running these commands in the same sequence with passport-saml's package.json on this repo, but still the command was failing for this repo and not for passport-saml.

kriss1897:
No, I didn't see any open npm tickets regarding the same. I found some old ones and some stackoverflow threads. Try this: https://www.google.com/search?q=cannot+find+module+npmlog+after+npm+update

cjbarth:
It seems that this has been [fixed](https://stackoverflow.com/a/34830277/271351). Could we revert the change and try again?

PR #10, opened 2021-07-01
Add support for a failed logout response

cjbarth:
@srd90 Has reported a bug with SLO in node-saml, passport-saml, [here](https://github.com/node-saml/passport-saml/issues/419). This adds support for returning a failed logout response to the IdP. Additional modifications are needed in passport-saml to correctly call node-saml to generate this response.

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -50,7 +50,7 @@ async function processValidlySignedPostRequestAsync(
   self: SAML,
   doc: XMLOutput,
   dom: Document
-): Promise<{ profile?: Profile; loggedOut?: boolean }> {
+): Promise<{ profile: Profile; loggedOut: boolean }> {
   const request = doc.LogoutRequest;
   if (request) {
     const profile = {} as Profile;
@@ -66,8 +66,8 @@ async function processValidlySignedPostRequestAsync(
       throw new Error("Missing SAML issuer");
     }
     const nameID = await self._getNameIdAsync(self, dom);
-    if (nameID) {
-      profile.nameID = nameID.value!;
+    if (nameID?.value) {
+      profile.nameID = nameID.value;
       if (nameID.format) {
         profile.nameIDFormat = nameID.format;
       }
@@ -88,7 +88,7 @@ async function processValidlySignedSamlLogoutAsync(
   self: SAML,
   doc: XMLOutput,
   dom: Document
-): Promise<{ profile?: Profile | null; loggedOut?: boolean }> {
+): Promise<{ profile: Profile | null; loggedOut: boolean }> {
   const response = doc.LogoutResponse;
   const request = doc.LogoutRequest;
 
@@ -197,7 +197,7 @@ class SAML {
     }
   }
 
-  _generateUniqueID() {
+  _generateUniqueID(): string {
     return crypto.randomBytes(10).toString("hex");
   }
 
@@ -360,7 +360,7 @@ class SAML {
     return stringRequest;
   }
 
-  async _generateLogoutRequest(user: Profile) {
+  async _generateLogoutRequest(user: Profile): Promise<string> {
     const id = "_" + this._generateUniqueID();
     const instant = this.generateInstant();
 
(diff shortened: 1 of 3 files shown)


cjbarth:
@mikkopiu, @markstos I am @-mentioning you because of your involvement with the above-mentioned SLO issue.

src/saml.ts
```diff
@@ -402,10 +402,25 @@ class SAML {
...
+
+    const failStatus = {
+      "samlp:StatusCode": {
+        "@Value": "urn:oasis:names:tc:SAML:2.0:status:Requester",
+        "samlp:StatusCode": {
+          "@Value": "urn:oasis:names:tc:SAML:2.0:status:UnknownPrincipal",
+        },
+      },
+    };
+
```
srd90, line 423: Now that `node-saml/node-saml`'s target is to be standalone SAML library without any relations to any particular nodejs frameworks (unlike `node-saml/passport-saml`) this PR could be merged regardless of what happens to PR https://github.com/node-saml/passport-saml/pull/619

`node-saml` module's client programs should be able to select status codes returned in `LogoutResponse` furthermore client programs should be forced to do so from day one after starting to use this module so that they have to really pay attention to proper `LogoutRequest` processing (of course most of those prorgams would just fallback to provoding status codes indicating `Success` regardless how well `LogoutRequest` handling is implemented but it would be those programs' internal problem/flaw).

So I suggest that change `success` parameter into tuple of statusCodes and merge this PR.
@cjbarth @markstos

cjbarth: What do you mean "tuple of statusCodes"? There are lots of status codes to pick from, so I can see how we might want to have arguments that allow a first and second level option. Is that what you mean?

https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-samlpr/96b92662-9bf7-4910-ab16-e1c28bce962b
https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf (Section 3.2.2.2)

srd90: >… so I can see how we might want to have arguments that allow a first and second level option. Is that what you mean?

Yes

cjbarth: I spiked some code to do this and noted as I was looking at all the status codes, that most of them wouldn't apply to this case. There really are only two conditions here the apply: logout success or logout failure. The failure will always be us (the `Responder`, and it will always be of type `UnknownPrincipal`. Do you see another status code that would be appropriate there? If not, it seems wise to keep the API simple and leave this as a `boolean`.

cjbarth: @srd90, I was wondering if you had any feedback on my last comment here.

srd90: @cjbarth If `node-saml` is used independently to implement SP (I am not talking about `passport-saml`) it could be implemented so that if IdP used for some reason e.g. POST binding and if SP is configured/implemented to support LogoutRequest only via redirect binding (or vice versa) it - SP which uses `node-saml` - could respond with `UnsupportedBinding` subcode.

So yes…I see at least one additional code which could be used.

src/saml.ts
```diff
@@ -68,8 +68,8 @@ async function processValidlySignedPostRequestAsync(
       throw new Error("Missing SAML issuer");
     }
     const nameID = await self._getNameIdAsync(self, dom);
-    if (nameID) {
-      profile.nameID = nameID.value!;
+    if (nameID.value) {
```
markstos, line 71: Why the change from using `nameID` to `nameID.value` here?

markstos (commented):
This looks good to me, but do we need the 11,000 line diff to `package-lock.json` if `package.json` wasn't modified in this commit?

It would also be nice squash the merge commits into one of the commits in this PR. 

I left one question where I didn't understand why a return value changed.

src/saml.ts
cjbarth: `self._getNameIdAsync` always returns an object with `.value`, so, in reality, what we want to do is check if `.value` has a value. That also allows us to remove the non-null assertion in the following line.

It follows the pattern directly after it with `nameID.format`.

PR #11, opened 2021-07-06
Feature: add facility in config to add `<Extensions>` element in SAML request

kdhttps:
## Issue
close https://github.com/node-saml/passport-saml/issues/602
Linked PR: https://github.com/node-saml/passport-saml/pull/607

## Description

Add support in `passport-saml` where it will allow admin to configure fully customise `<Extensions>` element and add it into SAML Request.

`<Extensions>` element are use to request custom attributes and setting. Its structure and values different as as IDP configurations and requests.

## Use cases
1.  Use `<Extensions>` element to request attributes/data. [More details, oasis-open.org SAML V2.0 Protocol Extension for Requesting Attributes per Request](http://docs.oasis-open.org/security/saml-protoc-req-attr-req/v1.0/saml-protoc-req-attr-req-v1.0.html)
2. Use `<Extensions>` element to request with some custom config. For Example: In [suomi.fi](https://palveluhallinta.suomi.fi/en/tuki/artikkelit/59116c3014bbb10001966f70) idp, its allows SP to request with custom language code by passing it `<Extensions>` element. [More details](https://palveluhallinta.suomi.fi/en/tuki/artikkelit/59116c3014bbb10001966f70)

## Changes

- `src/node-saml/types.ts` : added `extensions` optional variable. The type is `any` and it accept [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) type values because as per use cases and IDPs', there are many cases and each use cases have different request format/data. Please let me know what is your view on this.
-  `src/node-saml/saml.ts`: added it in `request`. As it has already XMLBuilder values so no need to do any processing.
-  Test cases added
-  Docs updated 
-  No Breaking Changes

# Checklist:

- [x] Issue Addressed
- [x] Link to SAML spec
- [x] Tests included
- [x] Documentation updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ const saml = new SAML(options);
 - `logoutCallbackUrl`: The value with which to populate the `Location` attribute in the `SingleLogoutService` elements in the generated service provider metadata.
 
 - **SAML Extensions**
-- `samlExtensions`: Optional, The SAML extension provides a more flexible structure for expressing which combination of Attributes are requested by service providers in comparison to the existing mechanisms, [More about extensions](https://docs.oasis-open.org/security/saml-protoc-req-attr-req/v1.0/saml-protoc-req-attr-req-v1.0.html). It accepts fully customize [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) type. There are many possible values for the `samlExtensions` element. You can use [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) to request attributes/data as per your need in SAML Request.
+- `samlExtensions`: Optional, The SAML extension provides a more flexible structure for expressing which combination of Attributes are requested by service providers in comparison to the existing mechanisms, [More about extensions](https://docs.oasis-open.org/security/saml-protoc-req-attr-req/v1.0/saml-protoc-req-attr-req-v1.0.html). There are many possible values for the `samlExtensions` element. It accepts fully customize [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) type.
 
 ```javascript
 // Example
--- a/README.md
+++ b/README.md
@@ -101,6 +101,25 @@ const saml = new SAML(options);
 - `additionalLogoutParams`: dictionary of additional query params to add to 'logout' requests
 - `logoutCallbackUrl`: The value with which to populate the `Location` attribute in the `SingleLogoutService` elements in the generated service provider metadata.
 
+- **SAML Extensions**
+- `samlExtensions`: Optional, The SAML extension provides a more flexible structure for expressing which combination of Attributes are requested by service providers in comparison to the existing mechanisms, [More about extensions](https://docs.oasis-open.org/security/saml-protoc-req-attr-req/v1.0/saml-protoc-req-attr-req-v1.0.html). It accepts fully customize [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) type. There are many possible values for the `samlExtensions` element. You can use [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) to request attributes/data as per your need in SAML Request.
+
+```javascript
+// Example
+samlExtensions: {
+  "md:RequestedAttribute": {
+    "@isRequired": "true",
+    "@Name": "Lastname",
+  },
+  vetuma: {
+    "@xmlns": "urn:vetuma:SAML:2.0:extensions",
+    LG: {
+      "#text": "sv",
+    },
+  },
+},
+```
+
 ### generateServiceProviderMetadata( decryptionCert, signingCert )
 
 As a convenience, the strategy object exposes a `generateServiceProviderMetadata` method which will generate a service provider metadata document suitable for supplying to an identity provider. This method will only work on strategies which are configured with a `callbackUrl` (since the relative path for the callback is not sufficient information to generate a complete metadata document).
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -352,6 +352,10 @@ class SAML {
       request["samlp:AuthnRequest"]["samlp:Scoping"] = scoping;
     }
 
+    if (this.options.samlExtensions != null) {
+      request["samlp:AuthnRequest"]["samlp:Extensions"] = this.options.samlExtensions;
+    }
+
     let stringRequest = buildXmlBuilderObject(request, false);
     // TODO: maybe we should always sign here
     if (isHttpPostBinding && isValidSamlSigningOptions(this.options)) {
--- a/src/types.ts
+++ b/src/types.ts
@@ -123,6 +123,7 @@ export interface SamlOptions extends Partial<SamlSigningOptions>, MandatorySamlO
 
   // extras
   disableRequestAcsUrl: boolean;
+  samlExtensions?: any;
 }
 
 export interface StrategyOptions {
--- a/test/samlRequest.spec.ts
+++ b/test/samlRequest.spec.ts
@@ -1,0 +1,119 @@
+import { SAML } from "../src/saml";
+import { FAKE_CERT, SamlCheck } from "./types";
+import * as zlib from "zlib";
+import * as should from "should";
+import { parseString } from "xml2js";
+
+const capturedSamlRequestChecks: SamlCheck[] = [
+  {
+    name: "Config with Extensions",
+    config: {
+      entryPoint: "https://wwwexampleIdp.com/saml",
+      cert: FAKE_CERT,
+      samlExtensions: {
+        "md:RequestedAttribute": {
+          "@isRequired": "true",
+          "@Name": "Lastname",
+        },
+        vetuma: {
+          "@xmlns": "urn:vetuma:SAML:2.0:extensions",
+          LG: {
+            "#text": "sv",
+          },
+        },
+      },
+    },
+    result: {
+      "samlp:AuthnRequest": {
+        $: {
+          "xmlns:samlp": "urn:oasis:names:tc:SAML:2.0:protocol",
+          Version: "2.0",
+          ProtocolBinding: "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST",
+          AssertionConsumerServiceURL: "http://localhost/saml/consume",
+          Destination: "https://wwwexampleIdp.com/saml",
+        },
+        "saml:Issuer": [
+          { _: "onelogin_saml", $: { "xmlns:saml": "urn:oasis:names:tc:SAML:2.0:assertion" } },
+        ],
+        "samlp:NameIDPolicy": [
+          {
+            $: {
+              "xmlns:samlp": "urn:oasis:names:tc:SAML:2.0:protocol",
+              Format: "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress",
+              AllowCreate: "true",
+            },
+          },
+        ],
+        "samlp:RequestedAuthnContext": [
+          {
+            $: { "xmlns:samlp": "urn:oasis:names:tc:SAML:2.0:protocol", Comparison: "exact" },
+            "saml:AuthnContextClassRef": [
+              {
+                _: "urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport",
+                $: { "xmlns:saml": "urn:oasis:names:tc:SAML:2.0:assertion" },
+              },
+            ],
+          },
+        ],
+        "samlp:Extensions": [
+          {
+            "md:RequestedAttribute": [
+              {
+                $: { isRequired: "true", Name: "Lastname" },
+              },
+            ],
+            vetuma: [
+              {
+                $: { xmlns: "urn:vetuma:SAML:2.0:extensions" },
+                LG: ["sv"],
+              },
+            ],
+          },
+        ],
+      },
+    },
+  },
+];
+
+describe("SAML request", function () {
+  function testForCheck(check: SamlCheck) {
+    return function (done: Mocha.Done) {
+      function helper(err: Error | null, samlRequest: Buffer) {
+        try {
+          should.not.exist(err);
+          parseString(samlRequest.toString(), function (err, doc) {
+            try {
+              should.not.exist(err);
+              delete doc["samlp:AuthnRequest"]["$"]["ID"];
+              delete doc["samlp:AuthnRequest"]["$"]["IssueInstant"];
+              doc.should.eql(check.result);
+              done();
+            } catch (err2) {
+              done(err2);
+            }
+          });
+        } catch (err3) {
+          done(err3);
+        }
+      }
+      const oSAML = new SAML(check.config);
+      oSAML.getAuthorizeFormAsync("http://localhost/saml/consume").then((formBody) => {
+        formBody.should.match(/<!DOCTYPE html>[^]*<input.*name="SAMLRequest"[^]*<\/html>/);
+        const samlRequestMatchValues = formBody.match(/<input.*name="SAMLRequest" value="([^"]*)"/);
+        const encodedSamlRequest = samlRequestMatchValues && samlRequestMatchValues[1];
+
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        const buffer = Buffer.from(encodedSamlRequest!, "base64");
+        if (check.config.skipRequestCompression) {
+          return helper(null, buffer);
+        } else {
+          return zlib.inflateRaw(buffer, helper);
+        }
+      });
+    };
+  }
+
+  capturedSamlRequestChecks.forEach(function (check) {
+    it(check.name, testForCheck(check));
+  });
+});


kdhttps:
I think @cjbarth we can start discussion from here https://github.com/node-saml/passport-saml/pull/607#issuecomment-864079774

cjbarth:
What do you think about taking the object as JSON or XML literal and then, in the ctor, making sure it can be converted to an `XMLBuilder` object for internal usage? That would allow a user to pass in what they like, but it means that we get strong typing and can validate that the XML structure is correct and throw an error while constructing if they passed in something bad.

kdhttps:
Thank you @cjbarth for quick response. Currently user can pass json which is directly accepted by `XMLBuilder`
```
samlExtensions: {
  "md:RequestedAttribute": {
    "@isRequired": "true",
    "@Name": "Lastname",
  },
  vetuma: {
    "@xmlns": "urn:vetuma:SAML:2.0:extensions",
    LG: {
      "#text": "sv",
    },
  },
},
```

So could you please describe, instead of above one what json/xml format you like to pass?
```
samlExtensions: {
     ????
}
```

cjbarth:
I'm not suggesting a different format. I'm suggesting that you adjust the ctor so that we validate the incoming data and throw if it doesn't conform to some basic checks. I'd really prefer not to blindly take whatever is put there and just add it to the outgoing SAML document.

srd90:
If I remember correctly extensions type in schema is ”sequence of any” so there isn’t much to check except that provided xml elements are valid. 

When those two example xml elements provided by @kdhttps would be added/injected to authn request generated by `node-saml` resulting authn request saml protocol message would not be valid because value for `md` namespace prefix is not defined (side note: it refers propably to `urn:oasis:names:tc:SAML:2.0:metadata` and `RequestedAttribute` sample is propably copied from some SAML Service metadata example).

`node-saml` module could for example validate generated authnrequest XML message against SAML Schemas just before it is delivered to IdP. This way erronous/non-valid XML issues would be more easily detected by application developers at application side instead of receiving some cryptic error message from IdP.

In fact validation of SAML messages against SAML schema is proposed by OWASP (although following cheat sheet seems to be written from receiver point of view): https://cheatsheetseries.owasp.org/cheatsheets/SAML_Security_Cheat_Sheet.html

FWIW: Here is also example of European Union’s eIDAS specification which also uses extensions (in service metadata and authn request):
https://ec.europa.eu/cefdigital/wiki/display/CEFDIGITAL/eIDAS+eID+Profile
and
`eIDAS SAML Message Format`
https://ec.europa.eu/cefdigital/wiki/download/attachments/82773108/eIDAS%20SAML%20Message%20Format%20v.1.2%20Final.pdf?version=3&modificationDate=1571068651727&api=v2
see chapter 6 from pdf to concrete examples.

BTW. @kdhttps if I interpreted current content of this PR correctly `samlExtensions` are fixed during SAML object creation (i.e. same values are used for each authnrequest). If I have understood correctly values at authn request’s extensions could change per authnrequest (e.g. vetuma language which is used to signal end user’s preferred language to be used at IdP side UI).

kdhttps:
> BTW. @kdhttps if I interpreted current content of this PR correctly samlExtensions are fixed during SAML object creation (i.e. same values are used for each authnrequest). If I have understood correctly values at authn request’s extensions could change per authnrequest (e.g. vetuma language which is used to signal end user’s preferred language to be used at IdP side UI).

No sure Mate, Currently I have a requirement to keep it fixed. But I can see in [suomi.fi docs](https://palveluhallinta.suomi.fi/en/tuki/artikkelit/59116c3014bbb10001966f70) that there are some other ways also to pass language element.

About schema validation, I think we don't have a fixed schema to validate it with SAML Request XML, as you can see there are many `saml extensions` possibilities. I have gone through the [owasp.org saml secuirty cheat sheet docs](https://cheatsheetseries.owasp.org/cheatsheets/SAML_Security_Cheat_Sheet.html), it is very generic and mostly about TLS and HTTPs network security. There is [some point about SAML Schema validation](https://cheatsheetseries.owasp.org/cheatsheets/SAML_Security_Cheat_Sheet.html#validate-signatures) but they didn't provide any schema, it is up to us to make the first schema and validate it with SAML Request XML but again the same situation **many saml extensions possibilities**. 

So I guess, the application developer has to manually verify saml request. Please let me know if I missed something and Let me know your suggestions.

Thank you, Team!

srd90:
@kdhttps as I said there isn’t any schema which could be used to validate content which is put to extensions block
> If I remember correctly extensions type in schema is ”sequence of any” so there isn’t much to check except that provided xml elements are valid

I should have used word well-formed instead of valid in that particular paragraph.

On third paragraph my proposal was to use validating XML parser to validate e.g. produced authnrequest. This would implicitly check that extensions element doesn’t contain something that would not pass authnrequest validation from XML document validation point of view (for example elements which use unbound namespace prefixes or malformed xml which are pretty much only validations that could be done to content of extensions element because it - the content of extensions - can seq).

SAML protocol schema is available e.g. from here https://docs.oasis-open.org/security/saml/v2.0/saml-schema-protocol-2.0.xsd

About enabling different extensions element content per request. IMHO if authnrequest extensions content is meant to be used to provide different content per authnrequest then node-saml library should enable it instead of forcing developers to instantiate new SAML object per request  (of course implementation could merge some values from stack configuration options but it should also allow real per request content). Not directly related to this case but passport-saml has already one ”half baked” per request thing see https://github.com/node-saml/passport-saml/issues/541#issuecomment-788111948

btw. you can use e.g. this https://www.samltool.com/validate_xml.php to debug authnrequests produced by node-saml. It shall produce error if extensions contains elements which namespace prefix is unbound / undefined (I am referring to this pull request’s README.md content and specifically to md:RequestedAttribute stuff at that example)

src/saml.ts
```diff
@@ -352,6 +352,10 @@ class SAML {
       request["samlp:AuthnRequest"]["samlp:Scoping"] = scoping;
     }
 
+    if (this.options.samlExtensions != null) {
+      request["samlp:AuthnRequest"]["samlp:Extensions"] = this.options.samlExtensions;
```
srd90, line 356: Position of `samlp:Extensions` is defined in SAML protocol schema.
```xml
…
    <complexType name="RequestAbstractType" abstract="true">                                                 
        <sequence>                                                                                           
            <element ref="saml:Issuer" minOccurs="0"/>                                                       
            <element ref="ds:Signature" minOccurs="0"/>                                                      
            <element ref="samlp:Extensions" minOccurs="0"/>                                                  
        </sequence>                                                                                          
        <attribute name="ID" type="ID" use="required"/>                                                      
        <attribute name="Version" type="string" use="required"/>                                             
        <attribute name="IssueInstant" type="dateTime" use="required"/>                                      
        <attribute name="Destination" type="anyURI" use="optional"/>                                         
        <attribute name="Consent" type="anyURI" use="optional"/>                                             
    </complexType>
…
    <complexType name="AuthnRequestType">
        <complexContent>
            <extension base="samlp:RequestAbstractType">
                <sequence>
                    <element ref="saml:Subject" minOccurs="0"/>
                    <element ref="samlp:NameIDPolicy" minOccurs="0"/>
                    <element ref="saml:Conditions" minOccurs="0"/>
                    <element ref="samlp:RequestedAuthnContext" minOccurs="0"/>
                    <element ref="samlp:Scoping" minOccurs="0"/>
                </sequence>
                <attribute name="ForceAuthn" type="boolean" use="optional"/>
                <attribute name="IsPassive" type="boolean" use="optional"/>
                <attribute name="ProtocolBinding" type="anyURI" use="optional"/>
                <attribute name="AssertionConsumerServiceIndex" type="unsignedShort" use="optional"/>
                <attribute name="AssertionConsumerServiceURL" type="anyURI" use="optional"/>
                <attribute name="AttributeConsumingServiceIndex" type="unsignedShort" use="optional"/>
                <attribute name="ProviderName" type="string" use="optional"/>
            </extension>
        </complexContent>
    </complexType>
…
```
source: https://docs.oasis-open.org/security/saml/v2.0/saml-schema-protocol-2.0.xsd


It looks like this implementation would place it into wrong place if developer has configured also e.g. `Scoping`

kdhttps: thanks mate, you are right. it is at wrong place. I'll fix it.

kdhttps:
> SAML protocol schema is available e.g. from here https://docs.oasis-open.org/security/saml/v2.0/saml-schema-protocol-2.0.xsd

About, XML Schema validation, I found [libxmljs](https://github.com/libxmljs/libxmljs) which provide facility to [validate xml again schema](https://github.com/libxmljs/libxmljs/wiki#validating-against-xsd-schema). I need your suggestion on this lib. May I use it for validation?

I can see, node-saml is using `xml2js`, `xmlbuilder` and `xmldom` but no one have validation feature. Iet me if you have any other option or use `libxmljs`??

Thank you!

srd90:
> I need your suggestion on this lib. May I use it for validation?

@kdhttps just in case this question was directed for me: if node ecosystem does not have native javascript xml library which understand xml schemas I would not introduce yet another xml library dependency (which needs native library). I would just let developers to make sure that they don't mess up with extensions and if they do then they just have to communicate with their IdP provider whats wrong with authnrequests syntax.

 For the record: I am not using passport-saml or node-saml and not going to use these. Checking out passport-saml every now and then is more like a hobby. So you should wait @cjbarth and @markstos etc. comments before proceeding with those content checks requested by @cjbarth

README.md
```diff
@@ -101,6 +101,25 @@ const saml = new SAML(options);
 - `additionalLogoutParams`: dictionary of additional query params to add to 'logout' requests
 - `logoutCallbackUrl`: The value with which to populate the `Location` attribute in the `SingleLogoutService` elements in the generated service provider metadata.
 
+- **SAML Extensions**
+- `samlExtensions`: Optional, The SAML extension provides a more flexible structure for expressing which combination of Attributes are requested by service providers in comparison to the existing mechanisms, [More about extensions](https://docs.oasis-open.org/security/saml-protoc-req-attr-req/v1.0/saml-protoc-req-attr-req-v1.0.html). There are many possible values for the `samlExtensions` element. It accepts fully customize [XMLBuilder](https://www.npmjs.com/package/xmlbuilder) type.
```
srd90, line 105: Since this PR is about enabling `Extensions` to various requests (which extend `samlp:RequestAbstractType`) supported by `node-saml` there is also `LogoutRequest` which may contain request specific `Extensions`.

```xml
…
    <complexType name="RequestAbstractType" abstract="true">                                                 
        <sequence>                                                                                           
            <element ref="saml:Issuer" minOccurs="0"/>                                                       
            <element ref="ds:Signature" minOccurs="0"/>                                                      
            <element ref="samlp:Extensions" minOccurs="0"/>                                                  
        </sequence>                                                                                          
        <attribute name="ID" type="ID" use="required"/>                                                      
        <attribute name="Version" type="string" use="required"/>                                             
        <attribute name="IssueInstant" type="dateTime" use="required"/>                                      
        <attribute name="Destination" type="anyURI" use="optional"/>                                         
        <attribute name="Consent" type="anyURI" use="optional"/>                                             
    </complexType>
…
    <complexType name="LogoutRequestType">                                                                   
        <complexContent>                                                                                     
            <extension base="samlp:RequestAbstractType">                                                     
                <sequence>                                                                                   
                    <choice>                                                                                 
                        <element ref="saml:BaseID"/>                                                         
                        <element ref="saml:NameID"/>                                                         
                        <element ref="saml:EncryptedID"/>                                                    
                    </choice>                                                                                
                    <element ref="samlp:SessionIndex" minOccurs="0" maxOccurs="unbounded"/>                  
                </sequence>                                                                                  
                <attribute name="Reason" type="string" use="optional"/>                                      
                <attribute name="NotOnOrAfter" type="dateTime" use="optional"/>                              
            </extension>                                                                                     
        </complexContent>                                                                                    
    </complexType>
…
```
source: https://docs.oasis-open.org/security/saml/v2.0/saml-schema-protocol-2.0.xsd

PR's current implementation skip/ignore `LogoutRequest`

kdhttps: nice catch @srd90 , we will add it in another PR

kdhttps:
> For the record: I am not using passport-saml or node-saml and not going to use these. Checking out passport-saml every now and then is more like a hobby. 

Great :100:  @srd90 

> So you should wait @cjbarth and @markstos etc. comments before proceeding with those content checks requested by @cjbarth

Sure, let's wait!!!

cjbarth:
I'm not a big fan of using libraries that require `node-gyp`; I'd rather see pure JavaScript solutions. When I'm talking about validation, I'm looking for two things, primarily:

1. A value that is passed in, which `node-saml` can't accept, should return a clear error.
2. A value which is known to be wrong, or at least obviously malicious, shouldn't be accepted; see rule 1.

I understand that rule 2 would be a little hard to determine without validating against some schema. I'm not opposed to a schema validation, in fact, I think that would yield some real long-term benefits to the system. However, I feel like there is an opportunity for some low-hanging fruit here, short of such a full validation system. If we simply check the incoming `<extension>` payload to make sure it is valid XML and that it contains the basic structure (test a few keys and attributes), I think we will hit a minimum acceptable level of fitness for use.

README.md
cjbarth: Another PR, or in another push to this PR?

markstos:
Thanks for working on this @kdhttps. I just had a customer ask if we supported this feature as well, so I'm also interested to see this get merged.

kdhttps:
@cjbarth 

About,
> 1. A value that is passed in, which node-saml can't accept, should return a clear error.

After some research I found that we can add one validation where element must have namespace like @srd90 maintained [here](https://github.com/node-saml/node-saml/pull/11#issuecomment-881025810). I am not finding any other validation check for `<extensions>` element. 

About,
> 2. A value which is known to be wrong, or at least obviously malicious, shouldn't be accepted; see rule 1.

I found [this XML Security Cheet sheet](https://cheatsheetseries.owasp.org/cheatsheets/XML_Security_Cheat_Sheet.html). Most problem should be handle by parser(at IDP side) and we are not parsing xml. we are just taking JSON from user and passes it to `XMLBuilder` to make XML request. But we need to restrict malicious data like [this](https://cheatsheetseries.owasp.org/cheatsheets/XML_Security_Cheat_Sheet.html#well-formed-document-to-well-formed-document-normalized). To restrict this, I didn't add any code as It is already handled by `xmlbuilder npm`. In fact, it is apply many validations like [this examples](https://cheatsheetseries.owasp.org/cheatsheets/XML_Security_Cheat_Sheet.html#applications-processing-malformed-data).

Please review code and let me know if you want me to add/change anything.

Thank you!!!

src/utility.ts
```diff
@@ -20,3 +21,61 @@ export function signXmlResponse(samlMessage: string, options: SamlSigningOptions
...
+ * Validate the XMLBuilder input JSON, wether it has proper namespece or not
+ * @param jsonXMLElementKey - Key of the element, need key to understand its namespace
+ * @param jsonXMLElement - Just JSON which we want to pass to xmlbuilder to make xml
+ * @returns boolean | JSONXMLSchemaError
+ */
+export function validateXMLNamespace(
+  jsonXMLElementKey: string,
+  jsonXMLElementValue: any
+): boolean | JSONXMLSchemaError {
+  jsonXMLElementKey = assertRequired(jsonXMLElementKey, `key should be define`);
```
cjbarth, line 58: s/define/defined

test/utility.test.ts
```diff
@@ -0,0 +1,232 @@
...
+  it("should throw TypeError when values are invalid", () => {
+    should(() => {
+      validateXMLNamespace("", undefined);
+    }).throw();
+    should(() => {
+      validateXMLNamespace("", undefined);
+    }).throw(TypeError);
+    should(() => {
+      validateXMLNamespace("", undefined);
+    }).throw("key should be define");
```
cjbarth, line 167: s/define/defined

test/utility.test.ts
```diff
@@ -0,0 +1,232 @@
...
+    should(() => {
+      validateXMLNamespace("test", {});
+    }).throw("test XML Element value should be object and not empty");
+  });
+
+  xmlElementSchemaNamespaceCheck.forEach((check) => {
```
cjbarth, line 173: I'm not a big fan of this pattern. I'd rather just see the outer "describe" block (XML Element Schema Namespace Check) with a inner "it" block with each test in the array. That makes it much easier to debug tests, and, in this case, would probably actually be fewer lines of code. Thoughts @markstos ?

.mocharc.json
```diff
@@ -1,10 +1,10 @@
...
   "diff": true,
-  "extension": "spec.ts",
+  "extension": ".ts",
   "package": "./package.json",
   "recursive": true,
   "reporter": "spec",
   "require": ["choma", "ts-node/register"],
-  "files": "test/**/*.spec.ts",
-  "watch-files": "test/**/*.spec.ts"
+  "files": "test/**/*.*.ts",
```
cjbarth, line 8: Why was `spec` removed?

cjbarth (changes requested):

test/utility.test.ts
markstos: I agree. It would be more clear not use the big array and format each each test in the standard way.

test/samlRequest.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+  {
+    name: "Config with Extensions",
+    config: {
+      entryPoint: "https://wwwexampleIdp.com/saml",
+      cert: FAKE_CERT,
+      samlExtensions: {
+        "md:RequestedAttribute": {
+          "@isRequired": "true",
+          "@Name": "Lastname",
+          "@xmlns:md": "urn:vetuma:SAML:2.0:metadata",
```
srd90, line 17: @kdhttps I think you meant to use namespace `urn:oasis:names:tc:SAML:2.0:metadata` for `md` prefix instead of `urn:vetuma:SAML:2.0:metadata` (because `urn:oasis:names:tc:SAML:2.0:metadata` would be inline with example you have written to README).

src/utility.ts
```diff
@@ -20,3 +21,61 @@ export function signXmlResponse(samlMessage: string, options: SamlSigningOptions
...
+  return result;
+}
+
+/**
+ * Validate the XMLBuilder input JSON, wether it has proper namespece or not
+ * @param jsonXMLElementKey - Key of the element, need key to understand its namespace
+ * @param jsonXMLElement - Just JSON which we want to pass to xmlbuilder to make xml
+ * @returns boolean | JSONXMLSchemaError
+ */
+export function validateXMLNamespace(
```
srd90, line 54: IMHO reinventing/reimplementing XML validation does more harm than good.

This function seems to check only that top level element has namespace defined but if top level element does not define default namespace then this function should check that child elements / subtrees have namespace defined etc.

furthermore sometimes namespace prefix might already be defined at enclosing element (i.e. at AuthnRequest element level).

So at the end of the day this partial validation implementation just adds complexity but fails to catch all namespace issues (which it claims to validate). I repeat myself by saying that if nodejs ecosystem does not have pure javascript xml validation implementation which could be applied to whole AuthnRequest and if `node-gyp` based libraries are out of question then it would be easier from long term maintainability point of view to just let `node-saml` library users to debug their xml extension related issues with IdP admins (of course most of the developers would use e.g. online validation tools to debug whats wrong with their authrequests).

cjbarth: This is a valid analysis. The best analysis of valid XML from our point of view, would be to simply check that the entire XML payload is valid (can be parsed by our XML tools), and that after it is parsed it contains an `<extensions>` element in the correct location.

While I don't want to be particular, it seems this would be as simple as trying to convert the JSON to XML and making sure that process doesn't fail. Then, using the XPath tools we already have to check to make sure that the `<extensions>` element is there with some expected attributes or children. This might not be possible during object construction, but we can certainly do it at run-time before we send the XML to anyone else.

If I'm looking at this from the wrong direction, perhaps we should just remove all validation and create a future work item to explore the entirety of XML validation. Honestly, any tool can be modified to present bad XML to an IdP. I'm just looking to alert the user of `node-saml` that they're doing something that isn't right so they can catch it and not have to rely on the IdP to tell them. But, maybe that is too much to ask or the wrong approach.

README.md
kdhttps: I think another PR

.mocharc.json
kdhttps: Because I added Unit test case. File name is `utility.test.js`. But I agree to not use `*`, I added specific type `test/**/*.[spec|test].ts`

test/utility.test.ts
kdhttps: @cjbarth just double checking, You want to keep this "it" block array in another "describe" block, right?

test/samlRequest.spec.ts
kdhttps: Yes, but I found that [saml-schema-protocol-2.0.xsd](https://docs.oasis-open.org/security/saml/v2.0/saml-schema-protocol-2.0.xsd) not really define rules to validate specific namespace value to specific element. It just check that there should be namespace attribute with value. 
For Example: 
```
<samlp:Extensions xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol">
      <md:RequestedAttribute xmlns:md="urn:vetuma:SAML:2.0:metadata" isRequired="true" Name="Lastname" />
      <vetuma xmlns="urn:vetuma:SAML:2.0:extensions">
        <LG>sv</LG>
      </vetuma>
</samlp:Extensions>
```
I checked with above data with this [XML check Tool](https://www.samltool.com/validate_xml.php). It allows this namespace. Please let me know if there are rules to check specific namespace value for elements.

srd90: Yes. Extensions content can be sequence of anything as long as it is well formed xml. My point was that since you choosed to select `RequestedAttribute` as one of the examples at the README.md and because `md:RequestedAttribute` happens to be defined in the specification that you referenced at README.md and because NS for `md:RequestedAttribute` happens to be `urn:oasis:names:tc:SAML:2.0:metadata` at that document you could keep using `RequestedAttribute` from the same namespace in the test case also.

src/utility.ts
srd90: >Then, using the XPath tools we already have to check to make sure that the <extensions> element is there with some expected attributes or children. This might not be possible during object construction, but we can certainly do it at run-time before we send the XML to anyone else.

@cjbarth you are aware that `ExtensionsType` allows anything?
```xml
    <complexType name="ExtensionsType">
        <sequence>
            <any namespace="##other" processContents="lax" maxOccurs="unbounded"/>
        </sequence>
    </complexType>
```
source: https://docs.oasis-open.org/security/saml/v2.0/saml-schema-protocol-2.0.xsd

I.e. from saml protocol layer point of view (and from protocol layer program blocks point of view such as `node-saml` or `passport-saml`) there aren't any "expected attributes or children" which existence/awailability `node-saml` could check on its own and on behalf of developer who is using `node-saml`.

test/samlRequest.spec.ts
kdhttps: Oh Got it ! Just fixed. Now it is `urn:oasis:names:tc:SAML:2.0:metadata`. I though you want me to perform some validation on namespace value. Thank you!

src/utility.ts
cjbarth: I understand that, but we can check to make sure that the bit of XML provided, in JSON form, is well-formed and has the correct opening tag...

Honestly, I feel like this is causing more problems than it is worth. I know debugging the XML coming out of `node-saml` is painful, so I wanted to make sure that we are doing due-diligence with the data we're working with. Perhaps we are better off with no validation what-so-ever, especially in light of https://github.com/node-saml/passport-saml/pull/551#issuecomment-894407649 and figuring out some other way to help the debugging experience along.

I'll adjust my opinion on this matter calling for a simple property to taken in the value and leaving it at that if @markstos agrees. The tests would then simply be making sure that the provided value gets injected in the outgoing XML at the correction location. Documentation would fill in the rest.

markstos: I'll defer back to @cjbarth. Simple sounds good to me.

cjbarth: In that case, lets not validate the incoming value for anything but being a valid object per your original idea @kdhttps . Sorry for the back-and-forth on this. @srd90 and @markstos provided valuable feedback and the iterative process sometimes works out this way.

test/utility.test.ts
cjbarth: Based on the new simplified spec, we'll just need a test to make sure that the payload gets interested in the correct place in the resultant XML and that a bad data type causes and error when constructing. In that case, we don't have to worry about many of these tests; they can be removed.

.mocharc.json
```diff
@@ -1,10 +1,10 @@
 {
   "diff": true,
-  "extension": "spec.ts",
+  "extension": ".ts",
```
cjbarth, line 3: Why was `spec` removed?

cjbarth (changes requested):

.mocharc.json
kdhttps: because of unit test file name ends with `.test.ts`, the file name is `utility.test.js`. Anyway, as we decided to remove validation and test. I'll set it back to `spec.ts`. Thank you!

src/utility.ts
kdhttps: @cjbarth removed validation and test cases

kdhttps: @cjbarth please review again and let me know if I miss anything

README.md
cjbarth: Just to be thorough, this [comment](https://github.com/node-saml/node-saml/pull/11/files#r683201959) indicates that you'll do another PR for the `LogoutRequest` version of Extensions. Does that mean that we should be using a more descriptive name instead of `samlExtensions` so there isn't a future name collision for that next PR?

test/samlRequest.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+        if (check.config.skipRequestCompression) {
+          return helper(null, buffer);
+        } else {
+          return zlib.inflateRaw(buffer, helper);
+        }
+      });
+    };
+  }
+
+  capturedSamlRequestChecks.forEach(function (check) {
```
cjbarth, line 124: This can all be collapsed since the array is of length 1. There is no need to do things this way. If that doesn't make sense, I can push to this branch later on with the refactor that I'm talking about. LMK.

src/types.ts
```diff
@@ -123,6 +123,7 @@ export interface SamlOptions extends Partial<SamlSigningOptions>, MandatorySamlO
 
   // extras
   disableRequestAcsUrl: boolean;
+  samlExtensions?: any;
```
cjbarth, line 126: Should be of type `object`.

src/saml.ts
```diff
@@ -265,6 +265,14 @@ class SAML {
       request["samlp:AuthnRequest"]["@AssertionConsumerServiceURL"] = this.getCallbackUrl(host);
     }
 
+    const samlExtensions = this.options.samlExtensions;
+    if (samlExtensions != null) {
```
cjbarth, line 269: If this is a `number` or `string` it will cause problems, especially if we try to ...spread it. We might as well do a `typeof` check here.

test/samlRequest.spec.ts
```diff
@@ -0,0 +1,127 @@
+import { SAML } from "../src/saml";
```
cjbarth, line 1: Would a test or two to check to make sure that the incoming data is of `object` type be reasonable?

cjbarth (changes requested):

test/samlRequest.spec.ts
cjbarth: Or rather, that an incoming type other than `object` will fail to construct instead with a sensible error instead of a compiler-generated error?

src/types.ts
kdhttps: `@typescript-eslint/ban-types` does not allow the type `object`.
```
Don't use `object` as a type. The `object` type is currently hard to use ([see this issue](https://github.com/microsoft/TypeScript/issues/21732)).
Consider using `Record<string, unknown>` instead, as it allows you to more easily inspect and use the keys.eslint@typescript-eslint/ban-types
```

src/saml.ts
kdhttps: agreed

src/types.ts
cjbarth: Sure, use that instead. That will help TypeScript users make sure they get at least the basics of what is supposed to go in this field.

test/samlRequest.spec.ts
kdhttps: I agree to check the type and it should be an object and throw type error with the message `samlExtensions should be Object`

README.md
kdhttps: I think right. we should make it more descriptive like `samlAuthnRequestExtensions`, what do you think @cjbarth ?

src/types.ts
kdhttps: Sure, Done

src/saml.ts
```diff
@@ -267,6 +267,9 @@ class SAML {
 
     const samlExtensions = this.options.samlExtensions;
     if (samlExtensions != null) {
+      if (!(typeof samlExtensions == "object")) {
```
cjbarth, line 270: Perhaps a bit of a nitpick, but I'm wondering why this construction which separates the `!` from the `==` instead of using `!=`.

test/samlRequest.spec.ts
kdhttps: Currently, there is one check only but In the future, there may possible of many checks. But I have no problem mate, it is all yours'. once you get time, please push your refactor code.

src/saml.ts
kdhttps: My bad :(, I am a little bit in a hurry. Adding fix :) thank you

test/samlRequest.spec.ts
kdhttps: Done, Please review it again. 

Also one more thing is remaining [here](https://github.com/node-saml/node-saml/pull/11#discussion_r686789096).

>Just to be thorough, this comment indicates that you'll do another PR for the LogoutRequest version of Extensions. Does that mean that we should be using a more descriptive name instead of samlExtensions so there isn't a future name collision for that next PR?


I would like to suggest `authnSAMLRequestExtensions` instead of `samlExtensions` field, @cjbarth  @srd90 let me know your thoughts

srd90: @kdhttps whats stopping you at implementing logout extensions in this same PR. If it would be implemented in this PR we would not have to try to see into the future what should have been taken into consideration when logout request extensions is introduced so that there would not have been any unneccesary breaking chances due late introduction of logout part.

kdhttps: @srd90 , agreed but I found that it is not really an `extensions` element in logoutRequest. Could you please check again https://docs.oasis-open.org/security/saml/v2.0/saml-schema-protocol-2.0.xsd? Maybe I missed something, please guide!

srd90: You missed that `LogoutRequestType` extends `RequestAbstractType`. `RequestAbstractType` type defines extensions element (see [this logout request related comment](https://github.com/node-saml/node-saml/pull/11#discussion_r671817007)).

Situation is same with `AuthnRequestType` (see [this authn request related comment](https://github.com/node-saml/node-saml/pull/11#discussion_r671007653)).

Some random side notes:

You have said that you have some customer who wants to pass parameters per request and furthermore that IdP is suomifi. suomifi document https://palveluhallinta.suomi.fi/en/tuki/artikkelit/59116c3014bbb10001966f70 says that language code is expected in AuthnRequest (see "Table 4" from that document) and in LogoutRequest (see "table 20" from that document).

About general usability of current impementation (from developer point of view). Assume that you have to pass e.g. 3 different request specific parameters and that each of them could have 3 different values and that each parameter's value is determined at the time authentication request is about to be issued. At the moment developer has two options:
* instantiate 3x3x2 (2 comes from authn+logout) different passport-saml instances and select the one which has correct combination request paramer values in passport-saml's configuration options because at the moment this pull request allows configuration of request parameters only at passport-saml stack instantion time.
* instantiate new passport-saml stack instance for each authentication and logout request with correct set of request specific parameters.

Alternative could to allow setting request parameres in authenticate function call also (more about this in [this comment](https://github.com/node-saml/node-saml/pull/11#issuecomment-881025810) at paragraph starting with sentence "About enabling different extensions element content per request. ...")

kdhttps: Hello @srd90,
Thank you for the information.

> You missed that LogoutRequestType extends RequestAbstractType. RequestAbstractType type defines extensions element (see this logout request related comment).

Gotcha, I will work on it. I will take `samlAuthnRequestExtensions` parameter name for SAML Authn Request and `samlLogoutRequestExtensions` parameter name for SAML Logout Request.

>You have said that you have some customer who wants to pass parameters per request and furthermore that IdP is suomifi.

No, For now just a one-time config but as per your observation, it looks like others may need different extensions as per request. I would like to cover this feature in the new PR. For now, let's add only the basics functionality.

kdhttps:
@srd90 @cjbarth added `samlLogoutRequestExtensions` feature, please review

test/tests.spec.ts
```diff
@@ -136,6 +136,96 @@ describe("node-saml /", function () {
...
+        });
+
+        const logoutRequestPromise = samlObj._generateLogoutRequest({
+          nameIDFormat: "foo",
+          nameID: "bar",
+        });
+
+        logoutRequestPromise
+          .then(function (logoutRequest) {
+            parseString(logoutRequest, function (err, doc) {
```
cjbarth, line 210: Use `parseStringPromise` following the pattern elsewhere in this PR and keep this all as promises instead of having to use `done`.

test/samlRequest.spec.ts
cjbarth: I agree, it seems wise that we add functionality in bit-sized pieces as people need it. This seems like a base that we can work with without any breaking changes to add the new feature. Is that your estimation too @kdhttps , that adding these additional features would result in no breaking changes?

src/saml.ts
```diff
@@ -376,13 +387,27 @@ class SAML {
           "@xmlns:saml": "urn:oasis:names:tc:SAML:2.0:assertion",
           "#text": this.options.issuer,
         },
+        "samlp:Extensions": {},
```
cjbarth, line 390: Why do you add the empty value here and then delete it later when not needed? The pattern you used previously just adds the value if you need it and skips the extra logic to delete if not needed.

cjbarth (changes requested):

src/saml.ts
kdhttps: it is because strick type of NameID here https://github.com/node-saml/node-saml/blob/71d9fe5ace94e19f1f5f288f1093a331a9758471/src/types.ts#L60

and `samlp:Extensions` element should be placed before NameID. There is other way also but I choose and Like to keep this one for safe type.

test/tests.spec.ts
kdhttps: done

kdhttps:
https://github.com/node-saml/node-saml/pull/11#discussion_r708432574
> I agree, it seems wise that we add functionality in bit-sized pieces as people need it. This seems like a base that we can work with without any breaking changes to add the new feature. Is that your estimation too @kdhttps , that adding these additional features would result in no breaking changes?

@cjbarth sorry, didn't get your point. agree on what? I am not seeing any breaking changes. Both features are optionally configurable.

test/samlRequest.spec.ts
kdhttps: Sorry not getting your point. I can't see any breaking changes, Both features are optionally configurable.

src/saml.ts
kdhttps: @cjbarth once you get time, please review it

PR #17, opened 2021-08-20
Forty/update xmldom

forty:
# Description

Update xmldom (and xml-crypto which also depends on xmldom too) to fix CVE-2021-32796

xml-encryption would also need to be updated if they ever update. See https://github.com/auth0/node-xml-encryption/pull/88


forty:
Updated the PR to update xml-encryption as well, has they merged to xmldom update

PR #1, opened 2021-05-17
Update README.md

cjbarth:
Simple update to the README file to express the new intent of this library.

PR #19, opened 2021-08-27
Split saml.ts - MR #1

forty:
As discussed in https://github.com/node-saml/passport-saml/discussions/629, I would like to refactor the code of saml.ts to make it easier to review and to work with.

My idea is to split it by feature. So for example we would have:
- saml/logout.ts
- saml/authorize.ts
- saml/assertion.ts
- ... (TBD)

In those files, I would like to avoid having SAML class (ie `this` or `self`) or SamlOption as parameter, as I feel it makes it hard to understand the input of the function. Instead, each function would take as input exactly what it needs to operate. I think this makes review much easier. It also often allow to have stricter types.

Since there is a lot of code to move, in order to bring more confidence that I'm not breaking anything, my strategy is to change the tests as little as possible (so I still leave some legacy functions which are tested to make sure tests are not broken - I will clean later once the split is done)

This is a first MR, to make sure you like this approach. I would rather like to do several smaller MRs rather than a big large one with the split, to avoid conflicting too much with other PRs if that works for you.

The only test I'm really changing (removing) is a test that called a function with a `string` parameter with `null`, which is not possible with out TS config. I know it's possible in JS, but I feel we should not start testing anything that TS can guarantee us, as not having to do this is the main benefit of TS IMO.

Thanks for letting me know what you think (@markstos and @cjbarth especially)


cjbarth:
@forty , I'm trying to land these PRs in order. Are you in a position to fix these merge conflicts so that I can review the code?

forty:
@cjbarth Sorry for the late response, the idea of rebasing this got me a bit lazy :sweat: 

It was a bit tough, but  I think I managed :tada: hopefully I have not reverted too many things. At least I had `npm run test` run on every single commit, and all it green, so that's a good sign.

The good news is that the number of conflict I got on saml.ts alone is itself a good reason to split that file :D

forty:
(And yes, 100% agree this should be split in several PRs, but the idea of this PR was to show the direction I wanted to go. I'll make smaller PRs for next ones)

src/saml/common.ts
```diff
@@ -0,0 +1,99 @@
+import * as querystring from "querystring";
+import { ParsedUrlQuery } from "querystring";
+import * as util from "util";
+import * as zlib from "zlib";
+import * as algorithms from "../algorithms";
+import { keyToPEM } from "../crypto";
+
+const deflateRawAsync = util.promisify(zlib.deflateRaw);
+
+export const getAdditionalParams = (params: {
```
cjbarth, line 10: Do you think that the signature of this function is now a little more difficult to use since it will no longer determine which additional params are needed based on the operation? Basically, this is now just a glorified object merging utility.

src/saml.ts
```diff
@@ -462,90 +373,50 @@ class SAML {
...
+      messageType,
+      additionalParameters,
+      privateKey: this.options.privateKey,
+      signatureAlgorithm: this.options.signatureAlgorithm,
     });
-
-    return target.toString();
   }
 
   _getAdditionalParams(
```
cjbarth, line 406: Ideally, this wouldn't need to be here at all. Is there a reason it wasn't completely factored out?

src/saml.ts
```diff
@@ -38,80 +36,15 @@ import {
...
+import { certToPEM, generateUniqueId } from "./crypto";
 import { dateStringToTimestamp, generateInstant } from "./datetime";
+import { getAdditionalParams, requestToUrl } from "./saml/common";
+import { generateServiceProviderMetadata } from "./saml/metadata";
+import { processValidlySignedPostRequest, processValidlySignedSamlLogout } from "./saml/logout";
 
 const inflateRawAsync = util.promisify(zlib.inflateRaw);
 const deflateRawAsync = util.promisify(zlib.deflateRaw);
 
-interface NameID {
```
cjbarth, line 47: I believe this is a type that `passport-saml` depends on. Did you also try loading this in as a dependency of `passport-saml` and seeing if `passport-saml` still passes all of its tests?

If you need help with that LMK.

cjbarth (changes requested):
Overall I like the direction this is going, however, it seems there may have been some refactors that either break or complicate things in unexpected ways. I look forward to this landing though, huge improvements, not just from the refactor, but also a clarification and simplification of style.

This will be the next PR to land (baring a major security issue), so there should be no more tedius merge conflict fixing. Sorry about that, and thanks for being a good sport.

src/saml.ts
forty: I'll check that, thanks

test/tests.spec.ts
```diff
@@ -2312,12 +2312,6 @@ describe("node-saml /", function () {
       afterEach(async function () {
         await samlObj.cacheProvider.removeAsync("_79db1e7ad12ca1d63e5b");
       });
-      it("errors if bad xml", async function () {
```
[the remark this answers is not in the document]
forty: I have no idea... probably a mistake, in a rebase or before. I have pushed a commit to put it back.

src/saml/common.ts
forty: > Basically, this is now just a glorified object merging utility

That was also my though. On the other hand, if you look at what is left in _getAdditionalParams, the only thing it does is a ternary on the operationAdditionalParams.

My plan was to eventually get rid of _getAdditionalParams completely (as mentioned in another comment, I initially planed to do it in another PR to let the test untouched for this big code change, but let me know if I should do it here) and instead call directly getAdditionalParams

I have the same thoughts for _requestToUrlAsync. I think I'm not fond of passing this "operation" parameter to select the behavior of the function. 

If we really want to avoid duplicating the logic of selection the right additional params in the function, think I would rather have a function that does just that, and directly use getAdditionalParams.

For example replace
```
this._getAdditionalParams(RelayState, operation, overrideParams)
```
by

```
    return getAdditionalParams({
      relayState,
      globalAdditionalParams: this.options.additionalParams,
      operationAdditionalParams: getAddtionalParamsForOperation(operation)
      overrideParams,
    });
```

With that said, getAdditionalParams is still a glorified object merging utility. I'd say that's fine, because I would find duplicating that code a bit too much, but that could be fine too. Let me know what you think.

forty: (I realize that this would makes some things a bit more verbose, but personally I think that verbose is not always a bad thing)

forty:
@cjbarth I answered to all your comments, I'll wait for your answers before pushing more changes. 

I plan to push updates at fixup commits which are ready for autosquashing so that you can easily look at the new things only (and I'll rebase --autosquash when we are all good). if you prefer that I autosquash as I go, let me know.

src/saml.ts
```diff
@@ -1302,155 +1170,57 @@ class SAML {
...
-      if (decryptedIds.length !== 1) {
-        throw new Error("Invalid EncryptedAssertion content");
-      }
-      return await promiseWithNameID(decryptedIds[0]);
-    }
-    throw new Error("Missing SAML NameID");
+    return await processValidlySignedPostRequest(doc, dom, this.options.decryptionPvk ?? null);
   }
 
   generateServiceProviderMetadata(
```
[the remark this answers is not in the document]
cjbarth: I'd like to get @markstos and possibly some other thoughts on this. I'm in favor of keeping it all together for the above-mentioned reasons. I see no value in a function that does work that would potentially produce an invalid or useless output. The `node-saml` library _is_ the lower-level library for `passport-saml`; I can't think of a valid use-case where it itself would need to yet again be split, even if just internally.

src/saml/logout.ts
```diff
@@ -0,0 +1,114 @@
...
+    if (sessionIndex) {
+      profile.sessionIndex = sessionIndex[0]._;
+    }
+    return { profile, loggedOut: true };
+  } else {
+    throw new Error("Unknown SAML request message");
+  }
+};
+
+export const processValidlySignedSamlLogout = async (
```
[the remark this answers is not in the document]
cjbarth: I think we have gone both ways on this one. Initially I was against the `Async` suffix, but it was needed for refactor reasons. Then I read some stuff about how having that suffix can make code review easier (very obvious if someone forgot an `await`). I also try to remember that others will be using this library, and there is a strong convention in APIs to use the `async` suffix.

Having said that, anyone using TypeScript wouldn't have this problem. Then again, not all those using this library would be using TypeScript. I think I'm 51% in favor of the suffix since it is very common to have a mix of callback and async functions in JavaScript code, and this makes the difference very obvious to any consumer or reviewer, no matter the IDE.

I think though that we should use the same convention over at `passport-saml` no matter what though. @markstos , do you have any thoughts on this?

src/saml/common.ts
cjbarth: I was actually going to comment similarly, verbosity isn't always bad. I really favor obvious code, especially with so many people coming in and out of this code all the time. We want to promote successful and bug-free code writing wherever possible. I personally would be comfortable seeing your suggesting in "by".

src/saml.ts
cjbarth: Sure, let's be safer for this massive PR.

src/saml.ts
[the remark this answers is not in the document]
markstos: I side with @cjbarth's viewpoint here.

src/saml/logout.ts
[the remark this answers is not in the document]
forty: I'll but the -Async back for now, so we can have this discussion separately

src/saml.ts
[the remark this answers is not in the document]
forty: I merged both function :+1:

src/saml.ts
forty: The compilation fail, but only due to an unrelated issue: the call to `getLogoutResponseUrl` is missing the `success` parameter that was added recently.

Otherwise things seem fine, tests pass.

src/saml/common.ts
forty: Should I do that in another PR? This change forces to update the tests which are testing _getAdditionalParams directly, so if we agree on not touching the tests in that PR, it's better to wait the next one.

src/saml/common.ts
```diff
@@ -0,0 +1,99 @@
...
+    samlMessage[k] = additionalParameters[k];
+  });
+  if (privateKey != null) {
+    // sets .SigAlg and .Signature
+    signRequest(samlMessage, privateKey, signatureAlgorithm);
+  }
+
+  const target = new URL(targetUrl);
+  Object.keys(samlMessage).forEach((k) => {
+    // TODO if we want to assume value are string, we should use
```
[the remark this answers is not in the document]
forty: Do you want me to do it in this PR? (I fine either ways)

forty:
I pushed some updates, and asked a few more questions. Sorry, I promise to make smaller PR next time :)

src/saml.ts
cjbarth: That seems like it shouldn't happen. I've just landed some new code in to `passport-saml`, so maybe that was the source of the mismatch? When I get some time, I'll pull this branch and verify myself for a second set of eyes.

src/saml/common.ts
cjbarth: I was thinking that we'd do the entire refactor and then leave a stub for `_getAdditionalParams`.

src/saml/common.ts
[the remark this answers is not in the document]
cjbarth: No, another PR please... this is more than enough for one PR. In fact, if it was easy to split this PR up in to multiple PR's I'd be for that :)

src/saml/logout.ts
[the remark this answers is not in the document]
cjbarth: FWIW, when I was reviewing the code more thoroughly, it seems that we sometimes use the Async suffix for functions that return Promises and other times don't, so we're internally inconsistent. Thus, no matter what decision we make, we'll have to have some code changes to unify our convention internally.

src/saml.ts
```diff
@@ -1302,155 +1173,29 @@ class SAML {
...
-      "@Location": this.getCallbackUrl(),
-    };
-
-    let metadataXml = buildXmlBuilderObject(metadata, true);
-    if (this.options.signMetadata === true && isValidSamlSigningOptions(this.options)) {
-      metadataXml = signXmlMetadata(metadataXml, this.options);
-    }
-    return metadataXml;
+    return generateServiceProviderMetadata({
+      ...this.options,
```
cjbarth, line 1194: I see how this is convenient, and probably not worth changing in this PR, but I'm in favor of being explicit about what we send; it makes refactoring and code analysis tools easier to use.

src/saml/common.ts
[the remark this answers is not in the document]
forty: I can absolutely split, I kept the commits well isolated.

I'll do that.

src/saml.ts
forty: Ah yes, I might have not rebased on very latest commit indeed. I'll retry on the split PRs

forty:
@cjbarth I split the PR. Only issue is that I have no idea how to target my other branch in this repository for the next PR.

Not a big deal, I ordered the PRs, let's do them one by one, and I'll rebase whenever one is merged.

cjbarth:
@forty , do you think that you'll be able to address these conflicts as we push to a goal of having all the outstanding PRs done by April for when Node 12 is deprecated?

cjbarth:
@forty I really wanted to help land this, but it seems that your changes have broken a test. I recognize that this wasn't that way before, however, updating dependencies, which didn't break any tests, resulted in this new behavior for your changes.

cjbarth:
@forty Are you still interested in following through on these, or should we just close them?

cjbarth:
See #147.

PR #20, opened 2021-08-30
Set AuthnRequestsSigned in SP metadata if configured for signing.

barryhagan:
# Description

This PR adds the AuthnRequestsSigned attribute to the SPSSODescriptor in SP metadata when the SP has been configured to sign requests.

Currently if both the `options.privateKey` is set, and a signing cert is passed to `generateServiceProviderMetadata`, the requests will be signed.  The metadata should be indicating this behavior with `AuthnRequestsSigned="true"`.

The SAML spec states in 2.4.4 (https://www.oasis-open.org/committees/download.php/56785/sstc-saml-metadata-errata-2.0-wd-05.pdf):
> AuthnRequestsSigned [Optional]
> Optional attribute that indicates whether the <samlp:AuthnRequest> messages sent by this
> service provider will be signed. If omitted, the value is assumed to be false. [E7]A value of
> false (or omission of this attribute) does not imply that the service provider will never sign its
> requests or that a signed request should be considered an error. However, an identity provider
> that receives an unsigned <samlp:AuthnRequest> message from a service provider whose
> metadata contains this attribute with a value of true MUST return a SAML error response and
> MUST NOT fulfill the request.

If requests will be signed, it would be better if this attribute was set to require the IdP to reject unsigned requests.  Effectively, this enables a more secure by default behavior.

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [ x]
- Tests included? [x ]
- Documentation updated? [N/A ]

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -460,7 +460,7 @@ class SAML {
     Object.keys(additionalParameters).forEach((k) => {
       samlMessage[k] = additionalParameters[k];
     });
-    if (this.options.privateKey != null) {
+    if (isValidSamlSigningOptions(this.options)) {
       if (!this.options.entryPoint) {
         throw new Error('"entryPoint" config parameter is required for signed messages');
       }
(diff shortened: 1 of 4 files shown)


cjbarth:
@markstos Would you consider this a breaking change? I'm all for more secure by default, and we'll be releasing `master` as a 4.0 next, so even that won't be a problem, but for documentation purposes...

markstos:
@cjbarth It seems like a breaking change and like becoming more spec compliant or even a bug fix.  

Yes, it may cause a problem if there's a case where some SP has their public key set up... but then isn't signing requests. If this starts to cause a failure by being more secure, it may highlight the issue and lead to a resolution with better security.

This feels analogous to changing the type for a function from any to the correct type. Yes, it could break for someone who was doing something weird and inadvisable, but the whole it's a move towards more correct behavior.

Put yet another way: this project has always strived to be spec-compliance. By making this change we being consistent with that documentation.

cjbarth (approved):

PR #22, opened 2021-09-13
fix(validatePostResponseAsync): support missing NotOnOrAfter in SubjectConfirmationData

bbaliguet:
# Description

Address a regression introduced by https://github.com/node-saml/node-saml/commit/54a1e0457f8e59a9b6947894072f7f9fb2bf8387. `NotOnOrAfter` and `NotBefore` are optional according to the specification (see https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf, `2.4.1.2 Element <SubjectConfirmationData>`). This minimal change returns a 0 value for `processMaxAgeAssertionTime`, compatible with `checkTimestampsValidityError` which already handles optional `notBefore` and `notOnOrAfter` attributes.

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [x]
- Tests included? [x]
- Documentation updated? [ ]

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -1423,6 +1423,9 @@ class SAML {
     notOnOrAfter: string,
     issueInstant: string
   ): number {
+    if (!notOnOrAfter) {
+      return 0;
+    }
     const notOnOrAfterMs = dateStringToTimestamp(notOnOrAfter, "NotOnOrAfter");
     const issueInstantMs = dateStringToTimestamp(issueInstant, "IssueInstant");
 
--- a/test/tests.spec.ts
+++ b/test/tests.spec.ts
@@ -1663,6 +1663,28 @@ describe("node-saml /", function () {
         });
       });
 
+      it("xml document with empty NotOnOrAfter time in SubjectConfirmationData should validate", async () => {
+        const unsignedXml =
+          '<samlp:Response xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" ID="R689b0733bccca22a137e3654830312332940b1be" Version="2.0" IssueInstant="2014-05-28T00:16:08Z" Destination="{recipient}" InResponseTo="_a6fc46be84e1e3cf3c50"><saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer><samlp:Status><samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/></samlp:Status>' +
+          '<saml:Assertion xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Version="2.0" ID="pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa" IssueInstant="2014-05-28T00:16:08Z"><saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer><saml:Subject><saml:NameID Format="urn:oasis:names:tc:SAML:2.0:nameid-format:transient">ploer@subspacesw.com</saml:NameID><saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer">' +
+          '<saml:SubjectConfirmationData Recipient="{recipient}" InResponseTo="_a6fc46be84e1e3cf3c50"/></saml:SubjectConfirmation></saml:Subject><saml:Conditions NotBefore="2014-05-28T00:13:08Z" NotOnOrAfter="2014-05-28T00:19:08Z"><saml:AudienceRestriction><saml:Audience>{audience}</saml:Audience></saml:AudienceRestriction></saml:Conditions><saml:AuthnStatement AuthnInstant="2014-05-28T00:16:07Z" SessionNotOnOrAfter="2014-05-29T00:16:08Z" SessionIndex="_30a4af50-c82b-0131-f8b5-782bcb56fcaa"><saml:AuthnContext><saml:AuthnContextClassRef>urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport</saml:AuthnContextClassRef></saml:AuthnContext></saml:AuthnStatement></saml:Assertion>' +
+          "</samlp:Response>";
+
+        const signingKey = fs.readFileSync(__dirname + "/static/key.pem");
+        const signingCert = fs.readFileSync(__dirname + "/static/cert.pem", "utf-8");
+        const signedXml = signXmlResponse(unsignedXml, { privateKey: signingKey });
+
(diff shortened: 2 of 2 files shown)


srd90:
@bbaliguet dunno if I'm interpreting specs correctly but...

AFAIK saml core spec provides "general guidelines" for SAML XML schema usage and https://docs.oasis-open.org/security/saml/v2.0/saml-profiles-2.0-os.pdf specifies with more detail per profile how to use stuff defined in saml core (quote from profiles document):
> ... Interoperability is achieved by constraining this flexibility when warranted by adhering to profiles that define how to use these elements with greater specificity than the generic rules defined by [SAMLCore]. ...

`passport-saml`/`node-saml` seems to follow (mostly) `4.1 Web Browser SSO Profile`.

`4.1.3.5 Identity Provider Issues <Response> to Service Provider` says (among other things):
>... The exact format of this HTTP response and the subsequent HTTP request to the assertion consumer service is defined by the SAML binding used. Profile-specific rules on the contents of the `<Response>` are included in Section 4.1.4.2. If the HTTP POST binding is used, the `<Response>` message is delivered directly to the service provider in this step. ...

`4.1.4.2 <Response> Usage` contains comments about content of `SubjectConfirmationData`:
>* The bearer `<SubjectConfirmation>` element described above MUST contain a `<SubjectConfirmationData>` element that contains a `Recipient` attribute containing the service provider's assertion consumer service URL and a `NotOnOrAfter` attribute that limits the window during which the assertion can be delivered. It MAY contain an `Address` attribute limiting the client address from which the assertion can be delivered. It MUST NOT contain a `NotBefore` attribute. If the containing message is in response to an `<AuthnRequest>`, then the `InResponseTo` attribute MUST match the request's `ID`.

Furthermore `4.1.4.3 <Response> Message Processing Rules` defines among other things:
>Regardless of the SAML binding used, the service provider MUST do the following:
>...
>* Verify that the `NotOnOrAfter` attribute in any bearer `<SubjectConfirmationData>` has not passed, subject to allowable clock skew between the providers.
>
>...

---

Below is `Response` from a `"xml document with empty NotOnOrAfter time in SubjectConfirmationData should validate"` test case in your commit https://github.com/Spotme/node-saml/commit/99505858e679ee63092639ac3308d176ddec1ebe#diff-d677334e0db2010c4f7c1a4c9d1426cbf9b83e870436c09d3dbcae2877a1af7c (reformatted for better readability):
```xml
<samlp:Response
    xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion"
    xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol"
    ID="R689b0733bccca22a137e3654830312332940b1be"
    Version="2.0"
    IssueInstant="2014-05-28T00:16:08Z"
    Destination="{recipient}"
    InResponseTo="_a6fc46be84e1e3cf3c50">
  <saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer>
  <samlp:Status>
    <samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/>
  </samlp:Status>
  <saml:Assertion
      xmlns:xs="http://www.w3.org/2001/XMLSchema"
      xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
      Version="2.0"
      ID="pfx3b63c7be-fe86-62fd-8cb5-16ab6273efaa"
      IssueInstant="2014-05-28T00:16:08Z">
    <saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer>
    <saml:Subject>
      <saml:NameID Format="urn:oasis:names:tc:SAML:2.0:nameid-format:transient">ploer@subspacesw.com</saml:NameID>
      <saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer">
        <saml:SubjectConfirmationData
            Recipient="{recipient}"
            InResponseTo="_a6fc46be84e1e3cf3c50"/></saml:SubjectConfirmation>
    </saml:Subject>
    <saml:Conditions
        NotBefore="2014-05-28T00:13:08Z"
        NotOnOrAfter="2014-05-28T00:19:08Z">
      <saml:AudienceRestriction>
        <saml:Audience>{audience}</saml:Audience>
      </saml:AudienceRestriction>
    </saml:Conditions>
    <saml:AuthnStatement
        AuthnInstant="2014-05-28T00:16:07Z"
        SessionNotOnOrAfter="2014-05-29T00:16:08Z"
        SessionIndex="_30a4af50-c82b-0131-f8b5-782bcb56fcaa">
      <saml:AuthnContext>
        <saml:AuthnContextClassRef>urn:oasis:names:tc:SAML:2.0:ac:classes:PasswordProtectedTransport</saml:AuthnContextClassRef>
      </saml:AuthnContext>
    </saml:AuthnStatement>
  </saml:Assertion>
</samlp:Response>
```

If I interpreted SAML specs correctly it should have `NotOnOrAfter` attribute in `SubjectConfirmationData` element:
```diff
@@ -22,6 +22,7 @@
       <saml:SubjectConfirmation Method="urn:oasis:names:tc:SAML:2.0:cm:bearer">
         <saml:SubjectConfirmationData
             Recipient="{recipient}"
+            NotOnOrAfter="...some xs:datetime value...see saml specs..."
             InResponseTo="_a6fc46be84e1e3cf3c50"/></saml:SubjectConfirmation>
     </saml:Subject>
     <saml:Conditions
```
and **if I interpreted specs correctly this pull request's test case is not testing spec compliant response**.

---

Random side notes not neccessarely related to this pull request:

If specs were interpreted correctly then the possiblity to switch off `passport-saml`'s `NotOnOrAfter` verification by configuring `acceptedClockSkewMs` to `-1` (see [passport-saml/README.md @ 04ee9415](https://github.com/node-saml/passport-saml/tree/04ee9415f7aad1f7763690b8e4711af6157c1985#config-parameter-details)) is against specification because `Response` processing rules say that it must be verified.

Some answer from stackoverflow which addresses also`NotOnOrAfter` values: https://stackoverflow.com/a/29546696

bbaliguet:
Hi @srd90 and thanks for the extended review. 

I indeed did not align this with the  ` https://docs.oasis-open.org/security/saml/v2.0/saml-profiles-2.0-os.pdf` section on `4.1 Web Browser SSO Profile`, but kept consistency with the existing code which handles `NotOnAfter` attributes as optional.

cjbarth:
We are preferring compatibility with the spec wherever possible @bbaliguet . Is there a specific need you're trying to address with this PR?

bbaliguet:
Hi @cjbarth ,

There is no specific need, when updating `passport-saml` a change of behaviour was noticed. If you check the related commit (https://github.com/node-saml/node-saml/commit/54a1e0457f8e59a9b6947894072f7f9fb2bf8387) I think having `NotOnAfter` not optional anymore is a non-intended side effect. 

Let's keep it this way then as it matches Web Browser SSO Profile specification.

cjbarth:
SGTM. We'd be happy to entertain a PR to adjust the code to not treat that value as optional in cases where that is where the spec requires.

PR #24, opened 2021-09-16
Add option to sign generated metadata

barryhagan:
# Description

As a follow on for #23, this PR adds the ability to sign generated metadata.   This is an added security measure when doing signing certificate rotations or other metadata sharing.

With this option enabled, IdPs that poll for metadata updates can now verify the metadata with the existing signing certificates it has configured before accepting new certs.  This prevents an attack vector from man-in-the-middle or compromised metadata hosting locations (assuming the signing cert private key was not also compromised).

# Checklist:

- Issue Addressed: [X]
- Link to SAML spec: [ ]
- Tests included? [X]
- Documentation updated? [X]

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ const saml = new SAML(options);
 - `disableRequestAcsUrl`: if truthy, SAML AuthnRequest from the service provider will not include the optional AssertionConsumerServiceURL. Default is falsy so it is automatically included.
 - `generateUniqueId`: optional function which will be called to generate unique IDs for SAML requests.
 - `scoping`: An optional configuration which implements the functionality [explained in the SAML spec paragraph "3.4.1.2 Element <Scoping>"](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf). The config object is structured as following:
-- `signMetadata`: if truthy, adds a signature to the generated Service Provider metadata. `privateKey` must be set to use this option.
+- `signMetadata`: if true, adds a signature to the generated Service Provider metadata. `privateKey` must be set to use this option.
 
 ```javascript
 {
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ const saml = new SAML(options);
 - `disableRequestAcsUrl`: if truthy, SAML AuthnRequest from the service provider will not include the optional AssertionConsumerServiceURL. Default is falsy so it is automatically included.
 - `generateUniqueId`: optional function which will be called to generate unique IDs for SAML requests.
 - `scoping`: An optional configuration which implements the functionality [explained in the SAML spec paragraph "3.4.1.2 Element <Scoping>"](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf). The config object is structured as following:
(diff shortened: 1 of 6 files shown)


barryhagan:
@cjbarth could I get a maintainer to review this PR?  I would like to get this in and then I can get #30 ready as well.

README.md
```diff
@@ -66,6 +66,7 @@ const saml = new SAML(options);
 - `disableRequestAcsUrl`: if truthy, SAML AuthnRequest from the service provider will not include the optional AssertionConsumerServiceURL. Default is falsy so it is automatically included.
 - `generateUniqueId`: optional function which will be called to generate unique IDs for SAML requests.
 - `scoping`: An optional configuration which implements the functionality [explained in the SAML spec paragraph "3.4.1.2 Element <Scoping>"](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf). The config object is structured as following:
+- `signMetadata`: if truthy, adds a signature to the generated Service Provider metadata. `privateKey` must be set to use this option.
```
cjbarth, line 69: I think I'd like to tighten up things as we go forward on this project to use `true` instead of truthy.

src/saml.ts
```diff
@@ -1408,7 +1408,12 @@ class SAML {
       "@Binding": "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST",
       "@Location": this.getCallbackUrl(),
     };
-    return buildXmlBuilderObject(metadata, true);
+
+    let metadataXml = buildXmlBuilderObject(metadata, true);
+    if (this.options.signMetadata && isValidSamlSigningOptions(this.options)) {
```
cjbarth, line 1413: Likewise, lets just test `=== true` as that better matches the types.

cjbarth (changes requested):

src/saml.ts
barryhagan: 👍

README.md
barryhagan: 👍

test/tests.spec.ts
```diff
@@ -383,62 +384,87 @@ describe("node-saml /", function () {
...
 
-      const samlObj = new SAML(samlConfig);
-      const signingCert = fs.readFileSync(__dirname + "/static/acme_tools_com.cert").toString();
+        const metadata = samlObj.generateServiceProviderMetadata(null, signingCert);
+        metadata.split("\n").should.eql(expectedMetadata.split("\n"));
 
-      const metadata = samlObj.generateServiceProviderMetadata(null, signingCert);
-      metadata.should.containEql('AuthnRequestsSigned="true"');
+        const dom = parseDomFromString(metadata);
+        samlObj.validateSignature(metadata, dom.documentElement, [signingCert]);
```
cjbarth, line 466: I understand this will fail if the signature isn't valid, so the test is valid. However, is it appropriate to also assert something positive we should be expecting from this? That would make it clear from just reading the test that it is doing the right thing. It also removes an implicit dependency on the workings of `validateSignature`. Example: `validateSignature` changes from throwing to returning `true` or `false`, then this test stops testing what we think it doesn't but will still pass.

Thoughts?

barryhagan: yeah, this needs to check the boolean return.  added that.

cjbarth (approved):

PR #25, opened 2021-09-23
Make Audience a required setting

cjbarth:
# Description

As discussed in https://github.com/node-saml/passport-saml/issues/137, it would be wise to require `audience` checking by default and force a consumer to disable it if they don't want audience checking. This PR accomplishes this.

There was a need to simply the `assertRequired` function, which caused a few related changes that aren't directly related to the functional or test changes of the related issue.

# Checklist:

- Issue Addressed: [ https://github.com/node-saml/passport-saml/issues/137 ]
- Link to SAML spec: [  N/A ]
- Tests included? [X]
- Documentation updated? [X]


srd90:
Regarding this pull request's this commit [d7b5b99](https://github.com/node-saml/node-saml/pull/25/commits/d7b5b9939b7980687f0458b8d73f4097ffc8db6b):
Some of the competitors defaults to validate that login response's audience restrictions contain SP's EntityID. In case of passport-saml that would be value of `issuer` provided to strategy as configuration option.

[Shibboleth SP 3](https://shibboleth.atlassian.net/wiki/spaces/SP3/overview)'s audience rule documentation says:
> … the SP will normally ensure that its own entityID is an allowable value without special setup…

source: https://shibboleth.atlassian.net/wiki/spaces/SP3/pages/2065334566/Audience+Rule

[Clever/saml2](https://github.com/Clever/saml2):
> * `audience` - (String or RegExp) — If set, at least one of the `<Audience>` values within the `<AudienceRestriction>` condition of a SAML authentication response must match. Defaults to `entity_id`.

source: https://github.com/Clever/saml2/tree/fca725bc329487d2ae6cf0fdff93b3bae23e3323#options

cjbarth:
@srd90 I appreciate your feedback. I've made the default `audience` equal to `issuer`.

cjbarth:
@srd90, I think this is ready to go now, and I'd love your feedback.

srd90:
@cjbarth IMHO https://github.com/node-saml/node-saml/pull/25/files view contains still lot of unrelated changes (e.g. `assertRequired` changes at numerous places for various options) so it is quite hard to spot this PR related changes.

I didn't notice related `audience` option documentation change/PR at `node-saml/passport-saml` documentation.

cjbarth:
@srd90 I had to change how `assertRequired()` worked in order to get things working correctly for checking for audience in the `ctor`. I can make a change to this PR to back all that out and just use a temporary duplicate utility function like `assertRequired2()` that does the right thing, and then follow with a refactor PR. I'll take care of that now.

As for the PR over at `passport-saml` for this change, your right, that is missing. I think we still need to figure out a way to unify the documentation. Thanks for the heads-up though.

src/saml.ts
```diff
@@ -131,7 +131,9 @@ class SAML {
       throw new TypeError("SamlOptions required on construction");
     }
 
-    const options = {
+    assertRequired2(ctorOptions.cert, "cert is required");
```
srd90, line 134: unrelated change at production code (i.e. not some randomish test code change)...see also similar change at line 158/161

src/saml.ts
```diff
@@ -146,6 +148,7 @@ class SAML {
       path: ctorOptions.path ?? "/saml/consume",
       host: ctorOptions.host ?? "localhost",
       issuer: ctorOptions.issuer ?? "onelogin_saml",
+      audience: ctorOptions.audience ?? ctorOptions.issuer ?? "unknown_audience", // use issuer as default
```
srd90, line 151: This would be much clearer if `issuer` would be modified to required option (i.e. user must provide non null string instead of defaulting `null/undefined` to `onelogin_saml`). 

If someone set `ctor.issuer = null` and doesn't set `audience` or sets `audience = null` result would be
```javascript
...
issuer: "onelogin_saml",
audience: "unknown_audience",
...
```
which is not what documentation says ("if audience is not provided it deafults to issuer").

test/tests.spec.ts
```diff
@@ -715,17 +715,118 @@ describe("node-saml /", function () {
...
+            "</samlp:Response>";
+          const base64xml = Buffer.from(xml).toString("base64");
+          const container = { SAMLResponse: base64xml };
+          try {
+            const samlObj = new SAML({
+              ...noAudienceSamlConfig,
+              issuer: "https://app.onelogin.com/saml/metadata/371755",
```
srd90, line 803: From test case readability point of view this cold be e.g.:
```javascript
issuer: "SP issuer value which doesn't match values at audience fields i.e. as-if IdP would not have added this SP's entityId to audience list"
```
current value seems to be IdP's own entityID.

test/tests.spec.ts
```diff
@@ -715,17 +715,118 @@ describe("node-saml /", function () {
...
+            const { profile } = await samlObj.validatePostResponseAsync(container);
+            expect(profile).to.not.exist;
+          } catch (err: unknown) {
+            expect(err).to.exist;
+            expect(err).to.be.instanceOf(Error);
+            expect((err as Error).message).to.match(/SAML assertion audience mismatch/);
+          }
+        });
+
+        it("if audience not provided, fail", async () => {
```
srd90, line 771: This is a bit of "twister". Documentation says that if `audience` is not provided value of audience defaults to value of options' `issuer`. More precise testcase description would be "if issuer and audience not provided, fail" (which is not exact definition either because issuer would default to "onelogin_saml" and "audience" would default to "unknown_audience" (see earlier comment).

IMHO This testcase would not be needed if ties to onelogin would be cut off and `issuer` would be changed to required argument (and audience would default to `issuer` if it is not defined).

src/saml.ts
cjbarth: I agree, and I had planned on a follow-up PR to address this. It appears that the cascading changes are making the reviewing harder, and I didn't want to add to it.

test/tests.spec.ts
cjbarth: I've updated the name of the test. I agree that the test won't be needed after the change to make `issuer` mandatory. I'll make sure to remove this test as part of that change.

test/tests.spec.ts
cjbarth: I agree with this readability improvement. Done.

cjbarth:
Thank you very much @srd90 for reviewing this. I've incorporated all of your excellent suggestions. I've also removed all the typing upgrades to make the impactful run-time changes easier to see. I will follow up next with a change to make `issuer` required once this lands. Feel free to see if my changes make sense to you.

cjbarth:
@markstos , I'd like to get this landed so that I can give attention to some other outstanding PRs and hopefully get them all landed so we can include them in a 4.0 release. I really appreciate @srd90 's review, and am comfortable landing having implemented his comments. Of course, I'd love your comments too, but if you're crunched for time, I can just land this and move on with some of the other outstanding PRs so we don't keep other contributors waiting too long.

markstos:
I'm late, but this made sense to me. It's generally better to be "secure by default" and let people explicitly reduce or disable security.

PR #29, opened 2021-10-13
CacheProvider interface

### LINKED ISSUE #28: [ENHANCE] cacheProvider type should be an interface

If you want to implement your own `cacheProvider` the type must match the included in-memory `CacheProvider` class.  This includes internals of the class (`cacheKeys` and `options`) since those are not marked private.

export an interface for `cacheProvider` rather than the concrete in-memory implementation.

stubbing out the props necessary for TypeScript to not complain.

barryhagan:
# Description

Fixes #28 so custom cache providers have a clean interface.

# Checklist:

- Issue Addressed: [x]
- Link to SAML spec: [n/a]
- Tests included? [n/a]
- Documentation updated? [x]


markstos:
@barryhagan My impression is that this is a rarely used feature and developers seem to generally on board with moving from callback interfaces to promises. We could support both interfaces during a transition, but it's a also pretty quick refactor if developers do need to adapt their callback code to promises. Shims like Bluebird's `asCallback()` can help with that.

PR #30, opened 2021-10-15
Set a unique ID value in generated metadata

barryhagan:
# Description

When generating SAML metadata, the ID created in the `EntityDescriptor` element is generated based on the issuer value.  The intent of the ID value is to be unique for a document so if you were to change properties of your SP, the different metadata versions should not have the same ID.   While it is unlikely that this ID is being used as a cache control mechanism, it should change if the metadata changes.  Most common implementations generate a new ID each time metadata is generated, and this change does the same.

SAML core spec **1.3.4 ID and ID Reference Values**:
> Any party that assigns an identifier MUST ensure that there is negligible probability that that party or
any other party will accidentally assign the same identifier to a different data object

# Checklist:

- Issue Addressed: [x]
- Link to SAML spec: [x]
- Tests included? [x]
- Documentation updated? [n/a]

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -1365,7 +1365,7 @@ class SAML {
         "@xmlns": "urn:oasis:names:tc:SAML:2.0:metadata",
         "@xmlns:ds": "http://www.w3.org/2000/09/xmldsig#",
         "@entityID": this.options.issuer,
-        "@ID": this.options.issuer.replace(/\W/g, "_"),
+        "@ID": this.options.generateUniqueId(),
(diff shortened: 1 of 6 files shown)


cjbarth (approved):

barryhagan:
@cjbarth thank you.  Can I get another beta build published to npm so I can pick this up?

cjbarth:
@zoellner, @barryhagan is asking for another beta release. We are getting close to a final 4.0 release; we just have some outstanding PRs awaiting review.

PR #33, opened 2021-12-01
[Split saml.ts #2] Move requestToUrl out of saml.ts

forty:
Split of this PR https://github.com/node-saml/node-saml/pull/19 -second commit

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -6,7 +6,6 @@ import { URL } from "url";
 import * as querystring from "querystring";
 import * as util from "util";
 import { InMemoryCacheProvider } from "./inmemory-cache-provider";
-import * as algorithms from "./algorithms";
 import { signAuthnRequestPost } from "./saml-post-signing";
 import { ParsedQs } from "qs";
 import {
@@ -38,9 +37,9 @@ import {
   validateXmlSignatureForCert,
   xpath,
 } from "./xml";
-import { certToPEM, generateUniqueId, keyToPEM, removeCertPEMHeaderAndFooter } from "./crypto";
+import { certToPEM, generateUniqueId, removeCertPEMHeaderAndFooter } from "./crypto";
 import { dateStringToTimestamp, generateInstant } from "./datetime";
-import { getAdditionalParams } from "./saml/common";
+import { getAdditionalParams, requestToUrlAsync } from "./saml/common";
 
 const inflateRawAsync = util.promisify(zlib.inflateRaw);
 const deflateRawAsync = util.promisify(zlib.deflateRaw);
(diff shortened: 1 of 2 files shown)


cjbarth:
Please resolve conflicts.

src/saml.ts
```diff
@@ -463,54 +440,38 @@ class SAML {
...
-      if (!this.options.entryPoint) {
-        throw new Error('"entryPoint" config parameter is required for signed messages');
-      }
-
-      // sets .SigAlg and .Signature
-      this.signRequest(samlMessage);
-    }
-    Object.keys(samlMessage).forEach((k) => {
-      target.searchParams.set(k, samlMessage[k] as string);
+    return await requestToUrlAsync({
```
cjbarth, line 466: Based on our previous discussion, I was expecting this to be the only call in this function. I expected to see the rest of these checks in the `requestToUrlAsync` function instead of in this `_requestToUrlAsync` function.

cjbarth (changes requested):
Much easier to review. Thank you. I think the comments that I've made are things that we've already discussed.

src/saml.ts
forty: Yes absolutely, that was my understanding too, I have just not did that change yet :+1:

src/saml/common.ts
```diff
@@ -21,3 +28,72 @@ export const getAdditionalParams = (params: {
     overrideParams ?? {}
   );
 };
+
+const signRequest = (
+  samlMessage: ParsedUrlQuery,
+  privateKey: string | Buffer,
+  signatureAlgorithm: string
+): void => {
+  const samlMessageToSign: ParsedUrlQuery = {};
```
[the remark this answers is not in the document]
forty: @cjbarth No this function can never be called with a null private key 

Existing code 
```
   if (isValidSamlSigningOptions(this.options)) {
     // (...)
      this.signRequest(samlMessage);
    }
     // (...)
  }
```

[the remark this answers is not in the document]
cjbarth: It seems that this block (https://github.com/node-saml/node-saml/pull/33/files#diff-f4250f7596524ccbb6e049bd857c45700ad6d5d9124479fa15c355b297f33e80L502-L504) was removed and not added back anywhere else. Same for this (https://github.com/node-saml/node-saml/pull/33/files#diff-f4250f7596524ccbb6e049bd857c45700ad6d5d9124479fa15c355b297f33e80L206) block.

src/saml.ts
cjbarth: No worries. I'm just trying to keep up with you :)

cjbarth:
@forty Unless I'm mistaken, the ball is in your court with this one.

cjbarth:
See #147.

PR #35, opened 2021-12-01
[Split saml.ts #4] Move some logout function out of saml.ts

forty:
Split of this PR https://github.com/node-saml/node-saml/pull/19 -fourth commit


cjbarth:
See #147.

PR #36, opened 2022-01-03
Added X509 certificate to KeyInfo X509Data, if passed through options

ganesha289:
# Description
As per issue - [https://github.com/node-saml/passport-saml/issues/585](url) - X509Certificate is not getting added to KeyInfo of AuthnRequest.
Added fix for this issue. If signingCert (service provider's certificate) parameter is passed in options, then X509Certificate will be added to AuthnRequest.

# Checklist:

- Issue Addressed: [https://github.com/node-saml/passport-saml/issues/585 ]
- Link to SAML spec: [ ]
- Tests included? [ No]
- Documentation updated? [No]


cjbarth (changes requested):
This looks reasonable, but it needs a README update and a test.

ganesha289:
> This looks reasonable, but it needs a README update and a test.

Updated PR with the changes.. Can you please review again..

ganesha289:
@cjbarth, Updated the PR, please review

ganesha289:
> I won't be able to get to it until later this week at the earliest.

@cjbarth, Can you please help with this PR..

cjbarth:
I have been without my computer for more than a month and only just got it back... long story, but it has put me out of the loop, and I need to catch back up. Lots of `node-saml` stuff is on my to-do list.

markstos:
> I have been without my computer for more than a month and only just got it back... long story, but it has put me out of the loop, and I need to catch back up. Lots of node-saml stuff is on my to-do list.

Yikes! Welcome back.

cjbarth:
@ganesha289 please fix the build before I review.

ganesha289:
> @ganesha289 please fix the build before I review.

@cjbarth ,

I am getting same build errors for master code too.

Initially it fails for "prettier-check" script,
image

When I run "prettier-format" script, that issue get resolved, but with so many file modifications,
image

After fixing "prettier-check" issue, 9 unit tests failing,
image
image

One more issue I see is Code scanning result -> CodeQL, but that is failing for existing function crypto.ts -> removeCertPEMHeaderAndFooter 

Can you please help..

ganesha289:
> > @ganesha289 please fix the build before I review.
> 
> @cjbarth ,
> 
> I am getting same build errors for master code too.
> 
> Initially it fails for "prettier-check" script, image
> 
> When I run "prettier-format" script, that issue get resolved, but with so many file modifications, image
> 
> After fixing "prettier-check" issue, 9 unit tests failing, image image
> 
> One more issue I see is Code scanning result -> CodeQL, but that is failing for existing function crypto.ts -> removeCertPEMHeaderAndFooter
> 
> Can you please help..

@cjbarth, I'm facing same issue with master code.. Can you please help..

ganesha289:
> @ganesha289 Please see the PR I made against your branch, this will fix the formatting issues and provide tests for this code change. https://github.com/ganesha289/node-saml/pull/1/files
> 
> I'm guessing you didn't have prettier using the correct config file and that caused line ending changes in test files.

@barryhagan , Thank you very much for fixing the formatting issues and adding test cases. I have approved the PR.. Merging is enabled now.. Should I proceed with merging the PR..?

barryhagan:
> @barryhagan , Thank you very much for fixing the formatting issues and adding test cases. I have approved the PR.. Merging is enabled now.. Should I proceed with merging the PR..?

Yes, that should get your PR ready for review.

ganesha289:
> > @barryhagan , Thank you very much for fixing the formatting issues and adding test cases. I have approved the PR.. Merging is enabled now.. Should I proceed with merging the PR..?
> 
> Yes, that should get your PR ready for review.

@barryhagan , @cjbarth , Thank you for the help in issue resolution. I have merged PR - https://github.com/ganesha289/node-saml/pull/1 - in to this one.. Please review..

ganesha289:
@cjbarth / @barryhagan, Can you please help with the review..?

barryhagan:
> @cjbarth / @barryhagan, Can you please help with the review..?

sorry @ganesha289, wish I could help, but I'm not a maintainer of this project.

ganesha289:
@cjbarth, It would be great help if you review the PR..

src/xml.ts
```diff
@@ -129,6 +130,17 @@ export const signXml = (
   if (options.signatureAlgorithm != null) {
     sig.signatureAlgorithm = algorithms.getSigningAlgorithm(options.signatureAlgorithm);
   }
+  if (options.signingCert != null) {
```
cjbarth, line 133: I don't see how in this code cert rotation is supported.

test/saml-post-signing-tests.spec.ts
```diff
@@ -301,4 +301,75 @@ describe("SAML POST Signing", function () {
...
     });
   });
+
+  it("should sign an AuthnRequest and include Signature KeyInfo", async function () {
+    const xml =
+      '<AuthnRequest xmlns="urn:oasis:names:tc:SAML:2.0:protocol"><saml2:Issuer xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion">http://example.com</saml2:Issuer></AuthnRequest>';
+    const signingCert = fs.readFileSync(__dirname + "/static/cert.pem").toString();
+    const result = signAuthnRequestPost(xml, { privateKey: signingKey, signingCert });
+    const doc = await parseXml2JsFromString(result);
+    doc.should.be.deepEqual({
```
cjbarth, line 311: We have moved away from `should` to `chai`.

test/saml-post-signing-tests.spec.ts
```diff
@@ -301,4 +301,75 @@ describe("SAML POST Signing", function () {
...
+                    DigestValue: [{ _: "wHDDyV7rEQ/AQLYeLgsEUXX+Zxw=" }],
+                  },
+                ],
+              },
+            ],
+            KeyInfo: [
+              {
+                X509Data: [
+                  {
+                    X509Certificate: [
```
cjbarth, line 356: Instead of checking for deep equal, why not just check for the keys you want to be present? Otherwise, this tests ends up testing for more than it is designed to and may fail if other code adds more features or fixes unrelated bugs at another time.

src/types.ts
```diff
@@ -2,6 +2,7 @@ export type SignatureAlgorithm = "sha1" | "sha256" | "sha512";
 
 export interface SamlSigningOptions {
   privateKey: string | Buffer;
+  signingCert?: string | string[] | null;
```
cjbarth, line 5: We don't use this pattern very much anymore. It makes the code cleaner if we don't have to correct the type and do type narrowing. We'd rather the user just pass the correct type. If array is supported, just make this `string[] | null`. However, it doesn't seem, from the code below, that an array actually brings any value.

cjbarth (changes requested):
Please be sure to pay attention to the linting errors when running `npm test`. We'd like to keep that number on a downward trend.

cjbarth:
@ganesha289 Sorry for the delay. I'll have more time over the next week or so to help you get this landed.

README.md
```diff
@@ -43,6 +43,7 @@ const saml = new SAML(options);
 - `audience`: expected saml response Audience (if not provided, Audience won't be verified)
 - `cert`: the IDP's public signing certificate used to validate the signatures of the incoming SAML Responses, see [Security and signatures](#security-and-signatures)
 - `privateKey`: see [Security and signatures](#security-and-signatures).
+- `signingCert`: the service provider's public signing certificate used to embed in AuthnRequest in order for the IDP to validate the signatures of the incoming SAML Request, see [Security and signatures](#security-and-signatures)
```
cjbarth, line 46: Would this be the same signing cert as used in `generateServiceProviderMetadata`? It appears that way, in which case, would it seem reasonable to modify `generateServiceProviderMetadata()` to take this value if the `signingCert` argument isn't provided?

ganesha289: Yes, this is same certificate used in generateServiceProviderMetadata(). So, I given same parameter name for this.

src/types.ts
ganesha289: Actually, this is the same certificate used in "generateServiceProviderMetadata()", so I followed same convention used in service provider metadata.
image

src/xml.ts
ganesha289: Actually in my change, we don't need array or certificate rotation, as we need to add latest certificate in AuthNRequest, but same certificate used in generate service provider metadata function, so I'm also following the same.. Should we reaname this parameter with different name? But that will be duplication then..

src/xml.ts
```diff
@@ -121,6 +122,17 @@ export const signXml = (
...
+    const certArray = Array.isArray(options.signingCert)
+      ? options.signingCert
+      : [options.signingCert];
+    const cert = removeCertPEMHeaderAndFooter(certArray[0]);
```
ganesha289, line 129: I'm using existing function here - "removeCertPEMHeaderAndFooter" - but this failing my UT for polynomial regular expression.. as shown below.. Any suggestions on this..?
image

src/types.ts
cjbarth: I thought that this might be the same. You'll note that in this function, the array actually brings value, in the function you've added, the array does nothing. I think some consideration might need to be given to how to unite these two concepts.

src/xml.ts
cjbarth: Interesting find. The problem is with these lines:

```TypeScript
export const removeCertPEMHeaderAndFooter = (certificate: string): string => {
  certificate = certificate.replace(/-+BEGIN CERTIFICATE-+\r?\n?/, "");
  certificate = certificate.replace(/-+END CERTIFICATE-+\r?\n?/, "");
  certificate = certificate.replace(/\r\n/g, "\n");
  return certificate;
};
```

Specifically the `-+` part. There is a suggestion on how to fix that [here](https://codeql.github.com/codeql-query-help/javascript/js-polynomial-redos/). Feel free to open a PR to address this. Something as simple as this can even land before this, and very quickly at that.

cjbarth: This has been resolved by https://github.com/node-saml/node-saml/pull/79. Please merge `master` in to your branch and have a look at code coverage.

cjbarth: I just did the merge for you. Have a look.

README.md
cjbarth: I've merged `master` into your branch because it has some changes to `generateServiceProviderMetadata` and I wanted to make sure that your code fit with them.

cjbarth (commented):
This approach looks very reasonable, however, I note from some [research](https://lists.w3.org/Archives/Public/public-xmlsec-maintwg/2007Jun/att-0068/DNAMES_Section4.4.4.htm) that `<X509Certificate>` seems to support multiple certificates, which would match our other usage of `signingCert`. If that is true, it seems reasonable to try to support that, which, admittedly, might require some work on a dependent library. Thoughts?

cjbarth:
@ganesha289 Do you have bandwidth over the next week or two so we can work together to get this PR landed so that it makes it in to 4.0?

ganesha289:
> @ganesha289 Do you have bandwidth over the next week or two so we can work together to get this PR landed so that it makes it in to 4.0?

@cjbarth, Yes, I'm available in this week.. Should I work on adding changes for this issue - https://github.com/yaronn/xml-crypto/issues/243?

cjbarth:
@ganesha289 , please do. It seems that @LoneRifle is quite active on GitHub, so if we get a PR up with some tests, he might be able to land it and push out a minor release (along with that [other PR](https://github.com/yaronn/xml-crypto/pull/242) which looks pretty useful). I'll help to review the code you make a PR for over at `xml-crypto` to help move things along as best I can.

LoneRifle:
> @ganesha289 , please do. It seems that @LoneRifle is quite active on GitHub, so if we get a PR up with some tests, he might be able to land it and push out a minor release (along with that [other PR](https://github.com/yaronn/xml-crypto/pull/242) which looks pretty useful).

Thanks for flagging this. In the excitement I pushed out 2.1.4 too early, which means that changes from @ganesha289 will be pushed out in a proposed 2.1.5 release. @cjbarth - I think that @yaronn might be active enough to grant you control of xml-crypto and the corresponding npm publish rights, so that you are not blocked on me.

cjbarth:
@ganesha289 , based on https://github.com/yaronn/xml-crypto/issues/232, would migrating to https://github.com/auth0/node-xml-encryption be a solution to this PR?

ghsyeung:
Hi @cjbarth, I just followed through the threads of issues and have a few questions and observations. First, thanks for @ganesha289 and your work.

Some observations 
- it doesn't seem like https://github.com/auth0/node-xml-encryption handles multiple certificates
  - https://github.com/auth0/node-xml-encryption/blob/e8df80cce4b0f1442fb14f526dc94b7a5bd0fe34/lib/xmlenc.js#L25 seems to take a similar approach (of stripping header and footer) for a single certificate
- potential merge/migration in https://github.com/yaronn/xml-crypto/issues/232 
  - I understand the premise to merge both libraries, but I see that would incur quite a bit of refactoring effort
  - also I am not certain how that would impact the current PR (would you mind to elaborate?)

Questions
- I understand from the specification/schema (the one you linked above) that `<X509Certificate>` supports multiple certificates and the intention is to support cert rotation.
  - However, if we already know the exact private key used to sign the auth request, is there any reason to attach multiple certificates along? 
  - From my understanding, `<X509Certificate>` supports multiple certificates doesn't mean you need to send all certs at once. Since we already know the exact key used, we should avoid sending extra certs because that would increase the payload size.

I am new to this context (working on a PoC to integrate `passport-saml` with a few different idps) so please feel free to point out my misunderstandings. I think this PR would be quite useful and would love to see it merged soon.

cjbarth:
@ghsyeung , I believe the value of sending multiple certs would be so that the IdP would know both certs so it will work before and after the rotation. However, I have no need of this feature, so I don't want to be insistent.

I'm not a stake-holder on this, so I don't want slow things down, however, it does seem reasonable to add this support via https://github.com/yaronn/xml-crypto/pull/249. Basically, the direction there is that we shouldn't be expanding `FileKeyInfo`, we should be using something different like `StringKeyInfo`. If a PR showed up at that project adding that support to fill this need, along with tests, I'll gladly approve it and update this project to make use of it.

If the PR is rejected, then I'll gladly take a PR against this project per `xml-crypto`'s README which showed how to roll-your-own `KeyInfoProvider` class.

You'll not at the discussion in https://github.com/yaronn/xml-crypto/pull/249 that there is no suite of tests for the `FileKeyInfo`. Since I don't know exactly what is needed, I'd be happy just to see tests that enforce the current behavior and we can take it from there if it doesn't work. Even better would be documentation that shows what the resultant XML should look like so we can have tests that match that.

cjbarth:
@ganesha289 I'm really close to a 4.0 release, this is the last outstanding PR. Do you think you'll have time to address this in the near future so as to not hold up a 4.0 release longer than necessary?

src/xml.ts
```diff
@@ -123,6 +124,15 @@ export const signXml = (
   if (options.signatureAlgorithm != null) {
     sig.signatureAlgorithm = algorithms.getSigningAlgorithm(options.signatureAlgorithm);
   }
+  if (options.signingCert != null) {
+    const cert = removeCertPEMHeaderAndFooter(options.signingCert);
+    /* istanbul ignore next */
+    sig.keyInfoProvider = {
```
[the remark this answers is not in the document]
cjbarth: Now that `xml-crypto` is part of `node-saml` and based on the discussion at https://github.com/node-saml/xml-crypto/issues/243, is there some other approach we might take here to keep such custom code out of here and in `xml-crypto`?

cjbarth:
@ganesha289 I'm preparing a v5 release. Would you like to work with me to get this landed for that release?

ganesha289:
> @ganesha289 I'm preparing a v5 release. Would you like to work with me to get this landed for that release?

Sure @cjbarth

cjbarth:
@ganesha289 , did you want to start by fixing the merge conflicts and then have a look at the https://github.com/node-saml/xml-crypto project and see if we can't use some of the new features there to add this feature there and use it here. Now that we have both projects in the same organization, we can put the code where it belongs very easily.

cjbarth:
@ganesha289 , can you give this another look over to see if, after the merge, things look good for you?

cjbarth (approved):

PR #37, opened 2022-01-28
Updates validateInResponseTo option to accept a string in addition to…

hluedeke:
… a boolean. Increases the functionality of validateInResponseTo to 'always', 'onlyIfPresent', and 'never'. See node-saml/passport-saml#329.

# Description

Thank you for taking the time to create a PR to support this community-maintained project. We are all volunteers here, so features are only added by contributers like you. Please also be patient while the maintainers work to review your proposed changes for quality and effacy.

Please include a _summary of the change_ and, where appropriate, _which issue is addressed_.

Your description should also include a _comment about the use-case_ for this change, including what systems you may be interfacing with that require this change. This will help future users know how this library is being used in the wild and how well supported it is.

If the change involves the SAML specification, please _include a link to the relavent part(s) of the SAML specificaiton_. Doing so speeds up the review process and helps the maintainers make sure that this project remains spec-compliant. You might start [here](https://www.oasis-open.org/standards#samlv2.0) to find the part of the specificaiton that relates to this PR.

_Please include tests._ Doing so will ensure that the changes made in this PR are not undone or otherwise corrupted by future changes.

# Checklist:

- Issue Addressed: [x]
- Link to SAML spec: [n/a]
- Tests included? [ x]
- Documentation updated? [x ]

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -202,6 +202,13 @@ class SAML {
     }
   }
 
+  private getValidationSetting(setting?: string | boolean | undefined): string {
+    if (setting === "never" || setting === "always" || setting === "onlyIfPresent") {
+      return setting;
+    }
+    return setting ? "always" : "never";
+  }
+
   private signRequest(samlMessage: querystring.ParsedUrlQueryInput): void {
     this.options.privateKey = assertRequired(this.options.privateKey, "privateKey is required");
 
@@ -234,7 +241,7 @@ class SAML {
     const id = this.options.generateUniqueId();
     const instant = generateInstant();
 
-    if (this.options.validateInResponseTo) {
+    if (this.getValidationSetting(this.options.validateInResponseTo) !== "never") {
       await this.cacheProvider.saveAsync(id, instant);
     }
     const request: AuthorizeRequestXML = {
@@ -891,13 +898,15 @@ class SAML {
   }
 
   private async validateInResponseTo(inResponseTo: string | null): Promise<undefined> {
-    if (this.options.validateInResponseTo) {
+    if (this.getValidationSetting(this.options.validateInResponseTo) !== "never") {
       if (inResponseTo) {
         const result = await this.cacheProvider.getAsync(inResponseTo);
         if (!result) throw new Error("InResponseTo is not valid");
         return;
-      } else {
+      } else if (this.getValidationSetting(this.options.validateInResponseTo) === "always") {
         throw new Error("InResponseTo is missing from response");
+      } else {
+        return;
       }
     } else {
       return;
--- a/test/tests.spec.ts
+++ b/test/tests.spec.ts
@@ -713,6 +713,61 @@ describe("node-saml /", function () {
         });
       });
 
+      it("supports 'onlyIfPresent' option for validateInResponseTo - does not validate when not present", async () => {
+        const requestId = "_a6fc46be84e1e3cf3c50";
+        const xml =
+          '<samlp:Response xmlns:saml="urn:oasis:names:tc:SAML:2.0:assertion" xmlns:samlp="urn:oasis:names:tc:SAML:2.0:protocol" ID="R689b0733bccca22a137e3654830312332940b1be" Version="2.0" IssueInstant="2014-05-28T00:16:08Z" Destination="{recipient}"><saml:Issuer>https://app.onelogin.com/saml/metadata/371755</saml:Issuer><samlp:Status><samlp:StatusCode Value="urn:oasis:names:tc:SAML:2.0:status:Success"/></samlp:Status>' +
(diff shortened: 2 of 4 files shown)


ghusse:
Hello,

I'm greatly interested by this change, because we need to support idp-initiated login flow as well as the standard SP-initiated flow.

How can I help?

markstos:
@hluedeke Thanks for the comment. My company will benefit from either PR-- we are currently setting `validateInResponseTo: false` in some cases where some form of "if present" would be more secure.

markstos (changes requested):
I've read through this PR and #40 and I like elements of both. If you or @ghusse  are up for some refinements, I'd like to accept this PR with some changes to incorporate some aspects of #40.  Here are the changes I'd like to see to this  PR: 

 1. In the docs, the possible option values are more clearly presented as a bulleted list, as in #40: https://github.com/node-saml/node-saml/pull/40/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L91
 2. The typing in #40 is stronger by using an enum to represent the possible values of the string. Update the code here to use an enum like #40 https://github.com/node-saml/node-saml/pull/40/files#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aR88
 3. As a nit, I think `ifPresent` as used in #40 is sufficiently clear compared to `onlyIfPresent`. Let's go with the shorter name.
 4. I like that this PR created a function to abstract away handling the various possible values. The name could be a clearer by conveying /which/ validation setting its getting, and the signature should be cleaned up by looking up `this.options.validateInResponseTo` internally instead of passing it in each time it's used. So the change becomes:

```javascript
// before 
this.getValidationSetting(this.options.validateInResponseTo)

// after (looks up this.options.validateInResponseTo internally)
this.willValidateInResponseTo()
```

Finally, I'd like someone to take a second look at the test coverage that #40 contributes. It's definitely /more/ tests added than here in #37, but it's not clear that's actually adding more /coverage/ or just re-testing the same code through more tests. If the extra tests don't seem to add extra value, I just assume go with the smaller amounts of tests added here.

Thanks!

hluedeke:
Ok, I can take a look at it at some point in the next week or so. Otherwise, if you'd prefer to do it @ghusse or need it sooner, you are welcome to take a crack at it.

src/saml.ts
```diff
@@ -202,6 +202,13 @@ class SAML {
     }
   }
 
+  private getValidationSetting(setting?: string | boolean | undefined): string {
```
srd90, line 205: Shouldn't you be calling this function also at line 1111 of version `269ea7c8` of node-saml's saml.ts
https://github.com/node-saml/node-saml/blob/269ea7c8f1abe9ede03d4c9228d5a150013e9e46/src/saml.ts#L1111-L1139
because value `never` is `truthy`.

As a random side note (not related to this pull request): Aforementioned implementation is surrounded by quite odd "goto label" implementation:
https://github.com/node-saml/node-saml/blob/269ea7c8f1abe9ede03d4c9228d5a150013e9e46/src/saml.ts#L1043-L1140 
but maybe referred code block contains some java-/typecsript magic which requires such implementation instead of just removing `getInresponseTo` label and `break getInResponseTo;` statements (disclaimer: I do not speak java-/typescript/node.js). @markstos

test/tests.spec.ts
```diff
@@ -713,6 +713,61 @@ describe("node-saml /", function () {
         });
       });
 
+      it("supports 'onlyIfPresent' option for validateInResponseTo - does not validate when not present", async () => {
```
srd90, line 716: IMHO This test description does not "add up". Why should result be non-valid (or "Exact match Invalid Signature, as inResponseTo should not reject" as it says at the end of testcase) if Response is missing `inResponseTo` fields and configuration is `onlyIfPresent`?

I bet this test case's SAML response was copy paste from some other test case and it used to have validly signed assertion and it - SAML response - was modified for this test case so that `inResponseTo` attributes were removed from response and assertion (leaving assertion's signature invalid). I.e. what this test case seems to be testing is that when `options.validateInResponseTo` has `truthy` value `onlyIfPresent` that this particular code block is bypassed successfully (which already has sort of "onlyIfPresent" functionality due xpath search returning possibly `falsy` or empty array IF there didn't happen to be any Reponse element with inResponseTo attrtibute at any namespace - `passport-saml`/`node-saml` is not using namespaces in xpath queries):
https://github.com/node-saml/node-saml/blob/269ea7c8f1abe9ede03d4c9228d5a150013e9e46/src/saml.ts#L748-L757
and function fails at this point:
https://github.com/node-saml/node-saml/blob/269ea7c8f1abe9ede03d4c9228d5a150013e9e46/src/saml.ts#L783-L785
i.e. it never tests this functionality (whis is located in `processValidlySignedAssertionAsync` which is executed after assertion signature validation):
https://github.com/node-saml/node-saml/blob/269ea7c8f1abe9ede03d4c9228d5a150013e9e46/src/saml.ts#L1109-L1139

As a randon side note (note related to this pull request): Just in case if you wonder - while you are browsing the code: `passport-saml`'s/ǹode-samls signature validation is a bit of a "mess". As long as Response has either (validly signed assertion) OR (validly signed top level signature AND validly signed assertion) attacker is free to alter content of Response's inResponseTo, issuer etc values any way as he/she wants (i.e. "outer" inReponse value can always be controlled by attacker under previously mentioned conditions). If SAML authentication response has only one signature and if that signature is associated to Response then and only then outer inResponseTo value is also covered by `passport-saml`/`node-saml` XML signature validation (along with the rest of the SAML authn response). I.e. this particular testcase happens to test (due to how XML response which is used as test material is signed) only parts of the message which can be modified by attacker (without getting caught by such action).  Luckily it seems that `passport-saml`/`node-saml` tries to verify that IF there was `inResponseTo` value at Response level (i.e. if attacker didn't remove it completely) and IF assertion level `inResponseTo` is not equal with Response level `inResponseTo` then response is discarded due obvious "response modification attempt". 

Background (definitely not related to this pull request):
Quite a few other SP implementations provide configuration options to make missing or invalid top level signature a "hard error". I.e. if person who is configuring `passport-saml`/`node-saml` powered SP knows that his/her IdP always provide top level signature he/she can configure service provider discard SAML authentication responses which doesn't have valid top level signatures. For more background information see e.g. https://github.com/node-saml/passport-saml/discussions/671#discussioncomment-2136937

hluedeke: You got it - this was a copy/paste from another test case. I don't know all the inner workings of this library - I only really understand the assertion piece that was discussed in https://github.com/node-saml/passport-saml/issues/329. I therefore did not feel confident writing a test from scratch - I simply wanted to cover the change in behavior using a test case that appeared to already cover an end-to-end scenario. Open to suggestions on changing it

hluedeke:
> Ok, I can take a look at it at some point in the next week or so. Otherwise, if you'd prefer to do it @ghusse or need it sooner, you are welcome to take a crack at it.

I wanted to give an update here - things have been quite a bit busier than I expected last week and this week, so I haven't had a chance to look at anything yet. It will be at least next week (possibly longer) until things slow down enough where I can take another look.

src/saml.ts
hluedeke: Noted - thanks

cjbarth:
Closing in favor of #40

PR #40, opened 2022-02-07
Allow to validate InReponseTo only if provided, to support IDP-initiated login

ghusse:
# Description

This PR is another proposition following #37 on the same subject, described on [passport-saml](https://github.com/node-saml/passport-saml/issues/329)

IDP-initiated logins using SAML do not send the value of `InResponseTo`. When allowing IDP-initiated logins on a system, we must allow this value not to be provided.

On the other hand, we also want to authorize SP-initiated logins, and validate the parameter `InReponseTo` for these requests.

This validation is supposed to avoid some sort of replay attacks, and we don't want to weaken the security of SP-initiated logins by allowing IDP-initiated logins.

This contribution is backward compatible with the current option `validateInResponseTo`, which allows boolean values before these change.

The option now also accepts
- `always` (equivalent to the old value `true`)
- `never` (equivalent to the old value `false`)
- `ifPresent` => the new value

When the value `ifPresent` is provided, then `InResponseTo` will be validated if it is sent in the response.

This way, the security is not weakened for SP-initiated logins.

[IDP-Inititated login](http://docs.oasis-open.org/security/saml/Post2.0/sstc-saml-tech-overview-2.0-cd-02.html#5.1.4.IdP-Initiated%20SSO:%20%20POST%20Binding|outline)

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [x]
- Tests included? [x]
- Documentation updated? [x]


markstos (commented):
I'm aligned with the idea of this. Could you say more about how it compares to the competing PR? Is it functionally equivalent but coded in a different style, or is a functional difference?

What are merits of selecting this PR over the other?

ghusse:
Hello @markstos ,

Thanks for your answer.

This PR is equivalent of the competing PR, except maybe for the backwards compatibility (it may be the case but I did not check). I did not use the proposed commit but I re-coded the same idea.

To be completely honest, I need this functionality in a project, and I wanted to make the maximum for the functionality to be accepted. I made this PR with a comprehensive description and added some tests to have the most chances for this functionality to be merged, because it seemed that they were absent in the other PR.

I would completely understand that you prefer merging the first PR, I just need one or the other in one project 😊

I can also help on the other PR if any help is needed.

ghusse:
My first choice would be to merge #37 because it's the original idea.

src/types.ts
```diff
@@ -122,7 +128,7 @@ export interface SamlOptions extends Partial<SamlSigningOptions>, MandatorySamlO
   signMetadata?: boolean;
 
   // InResponseTo Validation
-  validateInResponseTo: boolean;
```
cjbarth, line 125: This is going to land in a semver-major release, let's just do the break here get rid of the boolean; that will keep other parts of the code cleaner.

src/saml.ts
```diff
@@ -234,7 +235,7 @@ class SAML {
     const id = this.options.generateUniqueId();
     const instant = generateInstant();
 
-    if (this.options.validateInResponseTo) {
+    if (![false, ValidateInResponseTo.never].includes(this.options.validateInResponseTo)) {
```
cjbarth, line 238: I appreciate this might be easier to code, but negations are typically easy to miss when working with the code later and require more cognitive load. Let's keep the same pattern and leave it as an affirmative.

Side note: I'm glad to be getting rid of opportunities for truthy and falsey to be valid answers. I'd much rather see `=== true` or `=== false` for a security product, so that we are 100% clear what we are processing and what the consumer is sending (no chance of `"false"` being true).

cjbarth: Another option is do to as @markstos suggested [here](https://github.com/node-saml/node-saml/pull/37#pullrequestreview-887974520) and convert this to a function, perhaps `willValidateInResponseTo()`. In that case, negation wouldn't be as likely to get lost in all the syntax (though I'd still prefer to avoid negation where possible).

test/tests.spec.ts
```diff
@@ -1450,135 +1458,206 @@ describe("node-saml /", function () {
...
+      );
 
-        const { profile } = await samlObj.validatePostResponseAsync(container);
-        expect(profile!.nameID!.startsWith("UIS/jochen-work")).to.be.true;
-        const value = await samlObj.cacheProvider.getAsync(requestId);
-        expect(value).to.exist;
-        expect(value!).to.equal("2014-06-05T12:07:07.662Z");
+      [false, ValidateInResponseTo.never].forEach((validateInResponseTo) => {
+        describe(`when validateInResponseTo is set to ${validateInResponseTo}`, () => {
+          it("onelogin xml document with InResponseTo not in the cache should validate", async () => {
```
cjbarth, line 1636: Correct me if I'm wrong, but this is the only new test; the rest are just wrapped for the values of `ValidateInResponseTo` that apply to them, correct?

ghusse:
As there are no updates on the original PR #37, I updated this PR to add the suggestion from @markstos to use a function to check if `inResponseTo`  must be validated.

Regarding the tests I added, I just added some tests with the new options:
- When the validateInResponseTo is ` true or `always`  => check that the parameter is validated
- When it is `false`  or `never`  => check that the parameter is ignored
- When it is `ifPresent` => check that the parameter is validated when present and ignore when absent.

I did not implement new tests, I just made sure the existing ones that have been written for the values `true`  and ` false`  still pass with the new options too.

test/tests.spec.ts
cjbarth: Ok, if this is the only new test, then I think your tests look good.

cjbarth (changes requested):

test/tests.spec.ts
ghusse: Yes, you are right. Other modifications are just updates of existing tests to check that options are working with new values.

src/saml.ts
ghusse: I was working on these changes while you commented. I created this function as you can see in the new version.

src/types.ts
ghusse: I was hoping to see it in a minor version in order for it to be available faster through `passport-saml`  (because, in the end, I need this).

ghusse:
@cjbarth did you mark this PR as "Changes requested" because of your comment on the boolean option values to be removed? Because the only comment attached to your request is the one about tests, on which you replied yourself that they were OK.

src/types.ts
cjbarth: `node-saml` will only be released as a 4.0 release. `passport-saml` will be released with a breaking 4.0 based on `node-saml`. The currently `passport-saml`, 3.x, does not depend on `node-saml`. So, if you need it sooner, then this PR will have to land against the 3.x branch of `passport-saml`.

Having said all that, we are shooting to release v4 along with Node 14 in a few weeks.

ghusse: Ok then, I'm making the change

cjbarth:
> did you mark this PR as "Changes requested" because of your comment on the boolean option values to be removed?

Yes. We continue to try to get this project to be more and more explicit. Unless @markstos or the other maintains have another opinion, I don't see a need to support two ways of doing something as long as we document the change.

README.md
```diff
@@ -88,7 +88,10 @@ const saml = new SAML(options);
 ```
 
 - **InResponseTo Validation**
-- `validateInResponseTo`: if truthy, then InResponseTo will be validated from incoming SAML responses
+- `validateInResponseTo`:
+  - if `true`, `"always"`, then InResponseTo will be validated from incoming SAML responses
```
cjbarth, line 92: After your latest changes, this documentation needs to be updated to remove the option to use a boolean.

cjbarth (changes requested):

README.md
ghusse: Thanks, you're right, I forgot it

cjbarth (approved):

PR #43, opened 2022-03-08
Support multiple Assertion SubjectConfirmation

afurlane:
# Description

This fixes the bug related to multiple subject confirmation present in the assert as stated [here](http://docs.oasis-open.org/security/saml/v2.0/sstc-saml-approved-errata-2.0-cd-02.html) search for "E26: Ambiguities Around Multiple Assertions and Statements in the SSO Profile"

# Checklist:

- Issue Addressed: [https://github.com/node-saml/passport-saml/issues/339]
- Link to SAML spec: [[Saml core specification](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf )]
- Tests included? [ x ]
- Documentation updated? [ ]


src/saml.ts
```diff
@@ -1074,37 +1074,43 @@ class SAML {
...
-          : null;
-        confirmData =
-          subjectConfirmation && subjectConfirmation.SubjectConfirmationData
-            ? subjectConfirmation.SubjectConfirmationData[0]
-            : null;
-        if (subject[0].SubjectConfirmation && subject[0].SubjectConfirmation.length > 1) {
-          msg = "Unable to process multiple SubjectConfirmations in SAML assertion";
-          throw new Error(msg);
-        }
+        if (subject[0].SubjectConfirmation && subject[0].SubjectConfirmation?.length > 0) {
```
cjbarth, line 1077: This doesn't seem to be doing what the spec calls for. The spec calls for "If multiple assertions are included, then each assertion's <Subject> element MUST refer to the same principal." However, I see no check for that. What is more, it seems that by use of `Array.find()`, you are allowing _some_ `SubjectConfirmation` to be valid instead of checking that they are _all_ valid.

cjbarth (changes requested):
Please try to keep the linting errors on a downward trend. This code increases the number of linting errors.

src/saml.ts
afurlane: Sorry maybe I'm not so good at JS/TS for this code fix. But I'll try to do my best to fix as requested. Let me explain what the fix was meant for. I refer to this part "Verify that any assertions relied upon are valid in other respects. Note that while multiple bearer <SubjectConfirmation> elements may be present, the successful evaluation of a single such element in accordance with this profile is sufficient to confirm an assertion. However, each assertion, if more than one is present, MUST be evaluated independently."
That is related to [Issue 339](https://github.com/node-saml/passport-saml/issues/339) of passport-saml. I would also like to add a unit test for this change but I don't know how to get such a signing request. It's ok to use one from our internal test system?

cjbarth: I don't think I have any particular object to you adding a test from your internal system as long as the license is compatible with this project and your employer and the code owner is OK with it. I'll have a look at it when it shows up and provide comments.

cjbarth:
After taking a closer look at your code, I think it is appropriate. It isn't super thorough, but it preserves the same basic logic we had, but supports the spec for multiple `SubjectConfirmation`. It does, however, appear that after merging in `master` there are some failing tests. Could you please have a look at those?

BTW, if you are able to run `npm ci && npm test` locally, then, in most cases, the automated tests here will pass.

afurlane:
Ok, now I think I understand this project test a bit more. Test upgraded and should run as expected. I'm stuck with npm ci && npm test because it fails at prettier check.

cjbarth:
> fails at prettier check

To fix that run `npm run prettier-format`.

cjbarth:
Also, while you're fixing the linting, I would recommend reversing the changes to `package-lock.json`. Since you didn't update `package.json`, there should be no reason to touch `package-lock.json`.

test/tests.spec.ts
```diff
@@ -12,6 +12,7 @@ import assert = require("assert");
 import { FAKE_CERT, TEST_CERT } from "./types";
 import { signXmlResponse } from "../src/utility";
 import { parseDomFromString, xpath } from "../src/xml";
+import { Console } from "console";
```
cjbarth, line 15: Why is this line needed?

cjbarth:
Your new logic causes lots of problems for certain tests. It appears that your logic changes which `SubjectConfirmation` gets processed, which breaks some tests. If we are adding support for testing multiple subject confirmations, I wouldn't expect the existing behavior to change for when there are valid `SubjectConfirmation`. Before I look into this more, I'd like to know: what do you make of this?

test/tests.spec.ts
afurlane: > Why is this line needed?

I's a mistake, I beg you pardon for this. I'm so sorry to be in hurry and let this commit out. Yesterday I spent my time wondering why I was unable to run ci & test on the codebase. So I end up installing nvm and 14.x version of node in WSL on windows. My previous working environment was wrong. Node 16.x, windows and I got wrong results from test run. The same code base on the new environment breaks up only 20 tests instead of 25. I don't know why and will take a look on this too, but after fixing my errors and completing my task. This long answer is menta to be a good response to your question. But to be onest, the right one should be "poor man debugging".

afurlane:
Your're right, I'll fix it. Thak you for your help.

cjbarth:
I have reviewed your code several times, and I had the hardest time figuring out why your code was causing tests to fail. Until I found two other problems that were exacerbated by your new code, which takes lightly longer to run. I've created PRs to fix those problems: #59 and #60.

Once those lands, we can merge `master` in to this branch, and then I think I'll be able to do a final review on your code, with passing tests, to make sure everything is good, and we can get this landed. Thank you for sticking with this.

afurlane:
Thank to you Chris for your help. :)

cjbarth:
Ok, I've merged in all the code that was causing some problems with your new code. Please have another look at this and comment if it still all looks good to you.

afurlane:
I made a change in the test placement because it was in the wrong section.

afurlane:
Ok, now I'm ready for merging

src/saml.ts
```diff
@@ -1077,37 +1077,45 @@ class SAML {
...
             }
+          );
+
+          if (subjectConfirmation) {
+            subjectConfirmation = subjectConfirmation[0];
+            confirmData =
+              subjectConfirmation && subjectConfirmation.SubjectConfirmationData
+                ? subjectConfirmation.SubjectConfirmationData[0]
+                : null;
+          } else {
```
cjbarth, line 1117: The tests pass without this `else` block, so either we're missing a test, or something is wrong.

It seems like something might be wrong. In the old code, if `subject[0].SubjectConfirmation` can't be used `subjectConfirmation` is set equal to `null`. In this code, it isn't. This `else` block should probably be set to match line 1080, not line 1111.

It actually seems the course of wisdom to initialize `subjectConfirmation` and `confirmData` before the check for length, just as before, so that if length turns out to be 1, the old behavior is preserved. If the length is >1, we fall in to this new code. This would require that `subjectConfirmation` not be used with your `.find()` to hold the intermediate results.

In this way, you only have to add tests for new new condition you're affecting, which you've done. With the changes you've made here, I'd like to see some more tests assuring the existing behavior.

src/saml.ts
```diff
@@ -1077,37 +1077,45 @@ class SAML {
...
-
-            const subjErr = this.checkTimestampsValidityError(
-              nowMs,
-              subjectNotBefore,
-              subjectNotOnOrAfter,
-              maxTimeLimitMs
-            );
-            if (subjErr) {
-              throw subjErr;
+        if (subject[0].SubjectConfirmation && subject[0].SubjectConfirmation?.length > 0) {
```
cjbarth, line 1080: Can be simplified to `subject[0].SubjectConfirmation?.length > 0` in current versions of TypeScript. Otherwise, the old syntax can be used: `subject[0].SubjectConfirmation && subject[0].SubjectConfirmation.length > 0`. Either two checks or one `?`; not both.

cjbarth (changes requested):

src/saml.ts
afurlane: You're right; Probably I was drunk to write that piece of code. The working code should read like: "if there is any subjectconfirmation then filter them for valid timestamp and give back the resulting array for further validation". So the "else" part is really a useless piece of code. I really apologize for this.

src/saml.ts
afurlane: I know; I managed to escape this fate a lot of times, now it's clear to me, I have to LEARN JavaScript/TypeScript. MDN is not enough. Correction done.

src/saml.ts
```diff
@@ -1077,37 +1078,42 @@ class SAML {
...
+                    maxTimeLimitMs
+                  );
+                  if (subjErr === null) res = true;
+                }
+              }
+              return res;
             }
+          );
+
+          if (subjectConfirmation) {
```
cjbarth, line 1112: I was doing some more testing with this code and found that your previous way of doing this was just wrong. `Array.prototype.find()` doesn't return an array, so trying to index into `subjectConfirmation` with a `[0]` always returned `null`, which, always caused the following code to fail.

I also found that we don't have coverage for the following code, but that should be fixed in another PR.

cjbarth (commented):
Please have a look at some of the changes I've made to make sure they make sense to you. This PR never did sit right with me, and I finally figured out why. Sadly we don't have the code coverage to detect a problem as your code introduced here. I'd love it if you could help write some tests for the block starting on 1124.

test/tests.spec.ts
```diff
@@ -669,6 +669,51 @@ describe("node-saml /", function () {
...
+          issuer: "onesaml_login",
+          audience: false,
+        });
+
+        const { profile } = await samlObj.validatePostResponseAsync(container);
+        assertRequired(profile, "profile must exist");
+        expect(profile.nameID).to.equal("vincent.vega@evil-corp.com");
+      });
+
+      it("valid xml document with multiple SubjectConfirmation should validate, first is expired so it should take the second one", async () => {
```
cjbarth, line 694: While I agree with the premise, I don't see any assertions to guarantee that the second one was the one that was actually used.

src/saml.ts
cjbarth: I've added code coverage reporting so you can see how your code is doing.

cjbarth (changes requested):

test/tests.spec.ts
afurlane: I agree with you but unfortunately I have no way to verify which one was taken into account. At least as far as I can see.

cjbarth: I did it for you. You just had to make the second `SubjectConfirmationData` invalid, and then verify that the validation of it fails.

cjbarth (changes requested):
Still looking for, if possible 100% of your changes covered by tests.

src/saml.ts
```diff
@@ -1077,37 +1078,40 @@ class SAML {
...
+            confirmData = subjectConfirmation.SubjectConfirmationData
```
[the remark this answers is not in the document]
cjbarth: I've made this fix for you.

cjbarth:
It appears that codecov is mistakenly analyzing a line of code as the local run of `nyc` appears to cover all your changed line. What do you think?

afurlane:
To be true, I think I'm not so skilled in code covrage to understand which line of code is causing problems. Please give me some time to try catch this.

afurlane:
I should admit, I cannot find the code row overcovered by the run of nyc. I think is because I not so skilled in code coverage. I beg you pardon, but I'm still looking because I want to know.

cjbarth:
Honestly, I think we are good to land this code, even with that unchecked branch. I really think it is a problem with code analysis. And, even if that branch never gets tested, it is easy to see that the code will be just fine. There isn't a hidden case there. Thoughts?

afurlane:
Chris I agree with you. Thank you very much for your help, I really appreciate it.

srd90:
@cjbarth @afurlane based on very quick look to this pull request's changes it looks as if it would be possible bypass SubjectConfirmation's `NotOnOrAfter` validation and/or  subjectConfirmation's `validateInResponseTo` validation (even if in response to validation is set to `always`.

---

This is version of critical part of the code **before** this pull request landed:
https://github.com/node-saml/node-saml/blob/5d50faaaf173866e7270c114cb8e0f07a85489ad/src/saml.ts#L1076-L1157

If authentication response had at least one subject confirmation and if that subject confirmation's timestamp validation fails processing of whole authentication response would be stopped (due thrown exception):
https://github.com/node-saml/node-saml/blob/5d50faaaf173866e7270c114cb8e0f07a85489ad/src/saml.ts#L1112-L1120

and if timestamp validations were disabled but inrespnose to validation required then old version would try to validate existing subjectconfirmations inResponseTo value.

---

Now lets look situation **after** this pull request:
https://github.com/node-saml/node-saml/blob/4c7b82d7a55b18020644acd4612698282f68d8eb/src/saml.ts#L1076-L1153

Lets assume that `passport-saml` instance is configured so that
1. `NotOnOrAfter` validation is enabled and 
2. `validateInReponseTo` is set to `always`.

Furthermore lets assume that someone replays old (previously recorded) authentication response which has at least one subjectconfirmation to ACS callback endpoint. I.e. authentication response which is signed by same IdP that `passport-saml` instance uses (signature verification passes) but which - recorded response - might be minutes/weeks/months/years old.

AFAIK new version 
https://github.com/node-saml/node-saml/blob/4c7b82d7a55b18020644acd4612698282f68d8eb/src/saml.ts#L1103-L1109
would not stop processing with exception if timestamps are not valid (sidenote: there would be no visible reason at any logs if response's subject confirmation timesstamp validations failed).

It is possible that `subjectConfirmation`:
https://github.com/node-saml/node-saml/blob/4c7b82d7a55b18020644acd4612698282f68d8eb/src/saml.ts#L1091
is `falsy` after aforementioned `find` operation.

So even if  `validateInResponseTo` returns `truthy` (e.g. if value was `always`) subjectconfirmation's `inResponseTo` validation would be skipped because `subjectConfirmation` is `falsy`.
https://github.com/node-saml/node-saml/blob/4c7b82d7a55b18020644acd4612698282f68d8eb/src/saml.ts#L1121-L1152
and `passport-saml` would proceed to process content of assertion.

There is assertion's `conditions` timestamp validation:
https://github.com/node-saml/node-saml/blob/4c7b82d7a55b18020644acd4612698282f68d8eb/src/saml.ts#L1165-L1171
but one of the steps which should have failed (existing subjectconfirmation's `inResponseTo` validation) is already bypassed. NOTE: I haven't checked relation of subjectconfirmations and assertion's notonorafter values (i.e. e.g. is it possible that assertion's notonorafter could be further in the future than subjectconfirmations??)

One could say that because of this 
https://github.com/node-saml/node-saml/blob/4c7b82d7a55b18020644acd4612698282f68d8eb/src/saml.ts#L768-L777
at the beginning of authentication response processing that it would take care of `inResponseTo` validation.

But if there isn't response level signature (i.e. if only assertion is signed) that particular in reponseTo value is controllable by attacker. I.e. attacker could initiate login process and copy `inResponseTo` value to recorded authentication response's response level `inResponseTo` field and that another inResponseTo validation would be skipped due falsy subjectConfirmation value (if subjectconfirmation's timestamps were not valid).

First `inResposeTo` value is controllable also if IdP signed authentication response so that it has response and assertion level signatures (due how passport-saml cannot be configured to require valid response level signature i.e. it is satisfied if there exists at least valid assertion level signature).

IMHO if value of `subErr` would be checked at this line
https://github.com/node-saml/node-saml/blob/4c7b82d7a55b18020644acd4612698282f68d8eb/src/saml.ts#L1115
and if value is not null and code would throw that error then subject confirmation's timestamp errors would be reported visibly and processing of whole authentication response wold be stopped. Furthermore it would be easier to see that there isn't any obvious ways to skip validation of subject confirmation's  `inResponseTo`.

cjbarth:
That is an interesting observation. I went over this PR so many times and caught lots of other little things, but missed this one :( Thank for taking the time to look this over. I believe the problem can be rectified by adjusting how the immediately following code branched.

If the following two branches were removed, then the `validateInResponseTo` evaluates to `true`, then a proper evaluation would take place. Really, these two conditional checks shouldn't be there now that we are checking `mustValidateInResponseTo`.

https://github.com/node-saml/node-saml/blob/3d61e918b272a6bf32872b7b266c1244f9c4a87a/src/saml.ts#L1124-L1125

In fact, removing those two lines results in all the tests still passing _and_ better code coverage. Does that seem to you like it would address you very valid concern?

cjbarth:
> MHO if value of `subErr` would be checked at this line

In order for this to work, we'd have to check for `falsy` of `subjectConfirmation` _and_ check to see if _any_ of the items processed by `.find()` caused `subjErr` to be not equal to `null`. That seems reasonable as well, but it seems that we might not need to know specifically why an assertion failed. Really, if an assertion failed, we just throw it away and report that we couldn't find an assertion that matched `inResponseTo`, since, really, we wouldn't ever want to check against an invalid assertion.

srd90:
>https://github.com/node-saml/node-saml/blob/3d61e918b272a6bf32872b7b266c1244f9c4a87a/src/saml.ts#L1124-L1125
>In fact, removing those two lines results in all the tests still passing and better code coverage. Does that seem to you like it would address your very valid concern?

@cjbarth according to issuehttps://github.com/node-saml/passport-saml/issues/339 reporter:
>This was unexpected behavior for me since the [SAML core specification](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf) states that there can be "zero or more" subject confirmations.

I.e. authentication response might not have any subjectConfirmations. If your remove those lines `passport-saml` would effetively change spec to "one or more" (if mustvalidateInresponseTo is configured to always / ifpresent (which - ifpresent - processing is determined by existence of possibly attacker controllable response level inResponseTo value)..

Another sidenote regarding response level `inResponseTo`:
* if saml authentication response doesn't have any subjetconfirmations and if IdP has signed just assertion then mustvalidateInResponseTo == always can be bypassed by initiating authentication and copy pasting fresh in response to fresh enought recorded authentication response and 
* if saml authentication response doesn't have any subjetconfirmations and if IdP has signed response and assertion then mustvalidateInResponseTo == always can be bypassed by initiating authentication and copy pasting fresh in response to fresh enought recorded authentication response. `passport-saml` library user could be aware that his/her IdP signs whole response but `passport-saml` doesn't provide mechanism to require valid response level authentication (related issue #58).

cjbarth:
> effetively change spec to "one or more"

You are correct.

> Another sidenote regarding response level `inResponseTo`:

I was just typing a comment of something similar. There is actually one more case. What should be the behavior if the entire response isn't signed, the assertions are, and no `inResponseTo` is provided (which is a legal case in other places in the code)? Even with the fix I suggested, an invalid assertion, which previously would have caused an exception would now just be ignored because there is nothing to compare it to in the case of `ValidateInResponseTo.ifPresent`.

All this to say that my proposed fix won't work on many levels. Which means that we need some more tests to make sure that no one tries it :)

The only obvious way to handle this would be your proposal. If there are subject confirmations, and none of them pass, we need to throw then and not pass things down the line. @afurlane Would you be in a position to code something up including a test case that fails with the code the way it is now?

afurlane:
@srd90, @cjbarth I just finished looking at the comments, absolutely right. It's ok if I start to work also on issue #82?

cjbarth:
> It's ok if I start to work also on issue #82?

@afurlane please do.

PR #44, opened 2022-03-16
Upgrade xml-encryption to 2.0.0 (fixes audit issue)

tkalliom:
This PR resolves the node-forge issue reported by `npm audit`.

The interface of xml-encryption has not changed. The only differences
are the node-forge dependency being dropped and the required Node
version being incremented to 12 (already required by node-saml).

The implementation changes in xml-encryption mean that it now produces
errors from Node rather than node-forge.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3483,11 +3483,6 @@
         "is-stream": "^1.0.1"
       }
     },
-    "node-forge": {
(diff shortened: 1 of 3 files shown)


markstos (approved):
LGTM

cjbarth (approved):

PR #47, opened 2022-04-02
add an assertion to remove a linting warning

ghusse:
# Description

This is a small PR to remove some warnings

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [ ]
- Tests included? [ ]
- Documentation updated? [ ]

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -470,12 +470,16 @@ class SAML {
     additionalParameters: querystring.ParsedUrlQuery
   ): Promise<string> {
     this.options.entryPoint = assertRequired(this.options.entryPoint, "entryPoint is required");
+    const requestOrResponse = assertRequired(
+      request || response,
+      "one of request and response is required"
+    );
 
     let buffer: Buffer;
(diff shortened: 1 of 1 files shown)


src/saml.ts
```diff
@@ -470,12 +470,16 @@ class SAML {
     additionalParameters: querystring.ParsedUrlQuery
   ): Promise<string> {
     this.options.entryPoint = assertRequired(this.options.entryPoint, "entryPoint is required");
+    const requestOrResponse = assertRequired(
+      request || response,
+      "one of request and response is required"
```
cjbarth, line 475: nit: "one of request _or_ response is required"

cjbarth (changes requested):

src/saml.ts
cjbarth: Or better "Either request or response is required"

ghusse: fixed, thanks

cjbarth (approved):

PR #53, opened 2022-04-02
remove useless any type declaration

ghusse:
# Description

It is not necessary to specify the type of these variables, as ` readFileSync`  returns a `string`.

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [ ]
- Tests included? [ ]
- Documentation updated? [ ]

### DIFF
--- a/test/tests.spec.ts
+++ b/test/tests.spec.ts
@@ -2132,8 +2132,8 @@ describe("node-saml /", function () {
     });
   });
   describe("validatePostRequest()", function () {
-    const signingKey: any = fs.readFileSync(__dirname + "/static/key.pem", "ascii");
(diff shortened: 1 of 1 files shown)


cjbarth (approved):

cjbarth:
@ghusse You've been doing so much good work. If you wanted to try the refactor that @forty started in #19, I'd be happy to review that. Please, however, keep the pieces really small. The bigger the code is, the harder it is to buy out time to correctly review it.

cjbarth:
Even if you wanted to split the tests into smaller pieces (since you seem to be in the tests a lot), that would be a nice thing to have. Again, thanks for your help!

ghusse:
I'll take a look. Thanks for the suggestion. Maybe I'll work on unit tests.

I need to dig into the code a little bit further before starting this.

PR #54, opened 2022-04-02
Remove useless not null assertions

ghusse:
# Description

Small refactoring to remove a not null assertion that is not necessary

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [ ]
- Tests included? [ ]
- Documentation updated? [ ]

### DIFF
--- a/test/tests.spec.ts
+++ b/test/tests.spec.ts
@@ -662,9 +662,9 @@ describe("node-saml /", function () {
 
         const samlObj = new SAML({ cert: signingCert, audience: false, issuer: "onesaml_login" });
         const { profile } = await samlObj.validatePostResponseAsync(container);
-        const validatedProfile = assertRequired(profile, "profile must exist");
-        expect(validatedProfile.issuer).to.equal("https://evil-corp.com");
-        expect(validatedProfile.nameID).to.equal("vincent.vega@evil-corp.com");
+        assertRequired(profile, "profile must exist");
+        expect(profile.issuer).to.equal("https://evil-corp.com");
+        expect(profile.nameID).to.equal("vincent.vega@evil-corp.com");
         expect(profile).to.have.property("evil-corp.egroupid", "vincent.vega@evil-corp.com");
         // attributes without attributeValue child should be ignored
         expect(profile).to.not.have.property("evilcorp.roles");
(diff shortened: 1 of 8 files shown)


test/tests.spec.ts
```diff
@@ -678,8 +678,9 @@ describe("node-saml /", function () {
 
         const samlObj = new SAML({ cert: signingCert });
         const { profile } = await samlObj.validatePostResponseAsync(container);
-        expect(profile!.issuer!).to.equal("https://evil-corp.com");
-        expect(profile!.nameID!).to.equal("vincent.vega@evil-corp.com");
+        expect(profile).to.exist;
+        expect((profile as Profile).issuer).to.equal("https://evil-corp.com");
```
cjbarth, line 682: `as Profile` shouldn't be needed here if the return type of `validatePostResponseAsync()` is correct.

cjbarth (changes requested):

test/tests.spec.ts
ghusse: It seems that ` profile` is declared as nullable:

```ts
  async validatePostResponseAsync(
    container: Record<string, string>
  ): Promise<{ profile: Profile | null; loggedOut: boolean }> {
```

cjbarth: I see what you mean. There is a better way to do type narrowing on this. It appears that Chai doesn't do it. I have some other code that does. I'll get it up so you can see it.

ghusse: Can you share an example with me? @cjbarth

cjbarth: I already wrote code to do this, but it can't land just yet because of me not wanting to mess up #25. Here is what I've written, specifically a link to the function that will assert and do type narrowing all in one go: https://github.com/cjbarth/node-saml/blob/6f54cb690ca2fceae1044c7d2f2e6b226019ae1c/src/utility.ts#L4

ghusse: Ok, I used this method in the code, but I'm not a big fan of using some code that is external to the thing we are testing., but still belongs to the business code that is tested.

cjbarth: I'm not sure what you mean by code being external to the thing we're testing. Entire assertion libraries and libraries like `sinon` are external to the thing we are testing. In fact, they are also external to the all production code. At least this is used in our production code too.

cjbarth (approved):

PR #60, opened 2022-04-06
Correctly reset Sinon fake timers

cjbarth:
During some testing of the tests, it was found that some tests randomly failed. Diagnosis revealed it was because we were overwriting the `fakeClock` variable at times and therefore were not correctly `restore`ing the built-in timers, causing problems for following tests. This will fix that.

The pattern followed is to allow a `describe` blocks `beforeEach` and `afterEach` to set up and tear down fake timers if, and only if, each child `it` block needs the same fake timer. In every other case a `try...finally` block was used in the `it` block to build a local fake timer and tear it down, no matter what failures the test might experience.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2244,9 +2244,9 @@
       }
     },
     "@sinonjs/fake-timers": {
-      "version": "9.1.1",
-      "resolved": "https://registry.npmjs.org/@sinonjs/fake-timers/-/fake-timers-9.1.1.tgz",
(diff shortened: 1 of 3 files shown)


cjbarth:
This problem will be easier to find once [this](https://github.com/sinonjs/fake-timers/pull/426) is released.

cjbarth:
The entire method of this PR was changed because there is now an update to the Sinon library that will catch if you're using the fake timers incorrectly. Thus, the changes were paired back to be only what was needed. The code is still a little complex, but that can be saved for a proper test refactor.

PR #67, opened 2022-04-14
Extend available options for NameIDPolicy attributes

harrdou:
# Description

As per enhancement request node-saml/passport-saml#689, this adds two new options to specify the values for the `AlllowCreate` and `SPNameQualifier` attributes in the `<NameIDPolicy>` of authentication requests, as described in section 3.4.1.1 of [SAML-Core](https://www.oasis-open.org/committees/download.php/56776/sstc-saml-core-errata-2.0-wd-07.pdf).

This also changes the default behavior to include a `<NameIDPolicy>` with `AllowCreate="true"` as recommended in the above spec:

> Requesters that do not make specific use of this attribute SHOULD generally set it to “true” to
maximize interoperability. 

# Use-Case

These attributes support various use-cases relating to the use of persistent pairwise identifiers in the context of affiliations of service providers (ref: section 8.3.7 of the spec).

In our more advanced use-case, this feature is used by a proxying IdP (ref: section 3.4.1.5) to facilitate:
1. the replacement of IdPs that are leaving the federation, as well as
2. a large number of relying parties who are transitioning from SAML to OpenID Connect and need the pairwise subject identifiers of their users to be preserved.


README.md
```diff
@@ -60,6 +62,7 @@ const saml = new SAML(options);
 - `racComparison`: Requested Authentication Context comparison type. Possible values are 'exact','minimum','maximum','better'. Default is 'exact'.
 
 - `forceAuthn`: if set to true, the initial SAML request from the service provider specifies that the IdP should force re-authentication of the user, even if they possess a valid session.
+- `passive`: if set to true, specifies that the IdP must not visibly take control of the user interface and interact with the user.
```
cjbarth, line 65: I might be missing it, but I don't see a test for this.

src/saml.ts
```diff
@@ -278,14 +280,24 @@ class SAML {
       };
     }
 
+    const nameIDPolicy: XMLInput = {
+      "@xmlns:samlp": "urn:oasis:names:tc:SAML:2.0:protocol",
+    };
+
+    if (this.options.allowCreate != null) {
```
srd90, line 287: This pull request introduces `options.allowCreate` as (see addition at line 157):
```javascript
      allowCreate: ctorOptions.allowCreate ?? true,
```
and related addition at `types.ts` is at the moment:
```javascript
allowCreate?: boolean;
```
IMHO usage of nullish coalescing operator means that value of `this.options.allowCreate` shall always be `false` or `true` but never `null` at this point of code execution flow. I.e. this `if` statement would always evaluate to to `true` (and thus it would be "NOP" check).

README.md
harrdou: No test for `forceAuthn` as well, although only `passive` was absent from the documentation. I've added a new one for both.

harrdou:
Is there any reason that the `isPassive` option is passed to `generateAuthorizeRequestAsync()` as a parameter, while `forceAuthn` and other options are not?

Maybe clean this up for consistency?

src/types.ts
```diff
@@ -109,6 +109,8 @@ export interface SamlOptions extends Partial<SamlSigningOptions>, MandatorySamlO
   additionalParams: Record<string, string>;
   additionalAuthorizeParams: Record<string, string>;
   identifierFormat?: string | null;
+  allowCreate?: boolean;
```
cjbarth, line 112: Values that have a default shouldn't be optional (i.e. no `?`).

cjbarth (changes requested):
Please address the previous comments.

cjbarth (approved):
Thank you for your contribution!

Issue #58, opened 2022-04-04
Add option to fail hard if there should be valid top level signature and validation of top level signature failed


cjbarth:
Currently, if the validation of the top-level signature fails, we continue after setting a variable. We should have an option, or perhaps just fail, if we find a top-level signature that isn't valid. Right now, `node-saml` will allow an invalid top-level signature. This allowance allows for a whole class of crafty attacks that cause a top-level signature to be invalid, but which preserve the signature of an assertion. See https://github.com/node-saml/passport-saml/discussions/671 for a more thorough discussion of this matter.

Issue #82, opened 2022-05-05
[BUG] Potential bypass for `InResponseTo` Assertion Validation


cjbarth:
#43 introduced a bug whereby an attacker would modify a SAML document to effectively disable checking of assertion valid date/time ranges.

See https://github.com/node-saml/node-saml/pull/43#issuecomment-1118007252 for more information.

cjbarth:
@afurlane Are you available to work on this, or should we roll-back the bug-inducing PR until you get some time?

afurlane:
@cjbarth, sorry for being late. I'm working on the code, reading and re-reading the specs. I think I should be opening a pull-request soon.

afurlane:
@cjbarth I'm struggling with this trouble. If you could take a look, search for "valid xml document with multiple SubjectConfirmation should fail if no one is valid" in tests.spec.ts. Well, it is the same test in "valid xml document with multiple SubjectConfirmation should validate, first is expired so it should take the second one" and the assert reject checks for two differente messages. https://github.com/afurlane/node-saml/tree/bypass_for_inresponseto

afurlane:
@cjbarth, I found that:

`        await assert.rejects(samlObj.validatePostResponseAsync(container), (err) => {
          assert.strictEqual(err.message, 
            "No valid subject confirmation found among those available in the SAML assertion")
        });
`

Will work as expected. The test fails because of:
`1) node-saml /
       saml.js / 
         validatePostResponse checks /
           valid xml document with multiple SubjectConfirmation should fail if no one is valid:

      AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
+ actual - expected

+ 'InResponseTo does not match subjectInResponseTo'
- 'No valid subject confirmation found among those available in the SAML assertion'
      + expected - actual

      -InResponseTo does not match subjectInResponseTo
      +No valid subject confirmation found among those available in the SAML assertion`

Ok, so I think I need to fix the code and the tests too.

cjbarth:
Address by #87

PR #83, opened 2022-05-05
Add option to require a document signature.

cjbarth:
# Description

Addresses #58.


cjbarth:
@srd90, @markstos Do you think this option should be on by default forcing people to choose the insecure option?

src/saml.ts
```diff
@@ -158,6 +158,7 @@ class SAML {
       allowCreate: ctorOptions.allowCreate ?? true,
       spNameQualifier: ctorOptions.spNameQualifier,
       wantAssertionsSigned: ctorOptions.wantAssertionsSigned ?? false,
+      wantDocumentSigned: ctorOptions.wantDocumentSigned ?? false,
```
srd90, line 161: About naming...this is about signed authentication response. `Document` is quite generic term. SP may receive also `LogoutRequest` and `LogoutResponse` XML/SAML documents. Could it be `wantAuthenticationResponseSigned`?

Having said that I must also note that e.g. [tngan/samlify](https://github.com/tngan/samlify) seems to use `wantMessageSigned` (reference [tngan/samlify v2.8.4 docs/signed-saml-response.md](https://github.com/tngan/samlify/blob/9953b333ba2ab1613b671d0c2322db561895b3fd/docs/signed-saml-response.md))

src/saml.ts
```diff
@@ -782,6 +783,10 @@ class SAML {
         validSignature = true;
       }
 
+      if (this.options.wantDocumentSigned === true && validSignature === false) {
```
srd90, line 786: I do not use javascript/typescript so I have to ask this:

Lets assume this typescript library (`passport-saml`) is used inside application implemented with javascript.

Is it possible to configure `SamlStrategy` with
```javascript
...
wantDocumentSigned: "True",
...
```
and if so is it possible that even after all the code generated by typescript to javascript compiler (which may or may not introduce some type validations) this `"True"` ends up to `this.option.wantDocumentSigned`?

If it is possible (and since application developer clearly wanted to use value `true`) would it be safer to use this `if` statement:
```javascript
if (this.options.wantDocumentSigned !== false && validSignature !== true) {
```
This way all truthy (including `"True"`) values would be interpreted to "safer fallback" instead of unsafer.

srd90:
@cjbarth you asked:
>Do you think this option should be on by default forcing people to choose the insecure option?

Dunno. I do not have other opinion except that if you choose to default to forcing you may never switch it back to unsafer (i.e. e.g. if library users start to request unsafer default).

It seems that e.g. Shibboleth IdP and simplesamlphp (IdP) signs responses by default (dunno about other IdPs).
References:
https://shibboleth.atlassian.net/wiki/spaces/IDP4/pages/1265631694/SAML2SSOConfiguration
>The default value of signResponses for this profile is "true", in keeping with modern best practice. As long as one of the response or assertion are signed, use of the profile is "safe" in terms of authentication integrity, but there are vulnerabilities in XML Encryption that make signing responses advisable when the most common encryption algorithms are used. Some of the backstory around the signing defaults is discussed in this [thread](http://marc.info/?t=147134482600001&r=1&w=2).

https://simplesamlphp.org/docs/1.19/simplesamlphp-reference-idp-hosted
>saml20.sign.response
>    Whether <samlp:Response> messages should be signed. Defaults to TRUE . 
>    Note that this option also exists in the SP-remote metadata, and any value in the SP-remote metadata overrides the one configured in the IdP metadata. 

If you use non-forcing default you may want to think how well users of this library examine
1. SAML 
2. `passport-saml` and its documentation

before they start to use this library (i.e. are they happy when authentication starts to work or do they give extra push to verify that configuration is as safe as possible).

Lets take for example this: [node-saml/passport-saml v3.2.1 docs/adfs/README.md](https://github.com/node-saml/passport-saml/blob/6ba76ba3a015fea96a2dd38f661a6c1f85bc44a1/docs/adfs/README.md#create-a-passport-framework) which disables `NotOnOrAfter` validation (because `acceptedClockSkewMs` is set to `-1`) and lets see how widely this is used:
https://www.google.com/search?q=%22acceptedClockSkewMs%3A+-1%22+site%3Astackoverflow.com
https://www.google.com/search?q=%22acceptedClockSkewMs%3A+-1%22+site%3Agithub.com
there is enough hits to verify that first configuration which seems to work is used as-is.

or lets take for example: [node-saml/passport-saml v2.2.0 README.md#configure-strategy](https://github.com/node-saml/passport-saml/blob/0d983aa7806c4529efef75c7534cfdf9b7bd04a3/README.md#configure-strategy) which is the most unsecure example of `passport-saml` usage (prior to `3.0.0` that meant that all signature validation was disabled). Based on what has been discussed and referenced in issues related to `cert` option it is easy to say that there are too many `passport-saml` use places which just copy pasted that example without ever reading about [SAML security aspects](https://docs.oasis-open.org/security/saml/v2.0/saml-sec-consider-2.0-os.pdf) or without even reading `passport-saml`'s documentation (which said that it could be good idea to validate signature).

So...since SAML authentication response is considered safe if:
>...As long as one of the response or assertion are signed, use of the profile is "safe" in terms of authentication integrity...

(https://shibboleth.atlassian.net/wiki/spaces/IDP4/pages/1265631694/SAML2SSOConfiguration)
it means that current `passport-saml`'s signature validation is "safe" by default (yes there is that attacker controllable `inResponseTo` issue #82) but should `passport-saml` give extra push for safe by default and cause more "headaches" when projects are migrating to `4.0.0`??

src/saml.ts
cjbarth: I appreciate that this name is poor. I'm trying to find some SAML documentation to support having this option and see what it says. The closest that I've found is `WantAuthnRequestsSigned` and `authnRequestsSigned`, which doesn't appear to be the same.

src/saml.ts
cjbarth: That is a good idea. I'm not a big fan of using negations where possible, but it makes sense here. I've pushed up a PR to handle all such cases in the code-base: https://github.com/node-saml/node-saml/pull/85.

cjbarth:
After reading more about the spec and the defaults for signing in various parts of the spec, I'm of the opinion that it would probably be better to update the readme with some safer suggestions than it would be to set the default to sign. I'd like to hear @markstos opinion on this.

src/saml.ts
markstos: I agree a  good explicit name adds value, something like `wantAuthResponseSigned`.

markstos:
> After reading more about the spec and the defaults for signing in various parts of the spec, I'm of the opinion that it would probably be better to update the readme with some safer suggestions than it would be to set the default to sign. I'd like to hear @markstos opinion on this.

I'm strongly in favor of a "secure by default" approach. People tend to accept the defaults. If the defaults are insecure and then become popular, it could lead to a "class break" where many passport-samle projects are exploited at once.  As a breaking change, it can be clearly documented in a Changelog for a major version change, documenting what the change is how to restore the compatible, insecure behavior if necessary.

cjbarth:
@markstos I like the idea of making people pick the less secure option. I've changed this PR to force `wantAuthenticationResponseSigned` to be `true` by default.

I also noticed that we might want to do the same with `wantAssertionsSigned`. Thoughts?

Side note: does anyone here think that `wantAuthenticationResponseSigned` should be shortened to `wantAuthnResponseSigned`?

cjbarth:
> Based on new test's description it should be testing whether validation works if new `wantAuthnResponseSigned` is set explicitly to `true` (which is actually current default if not configured to false).

@srd90 , you are entirely correct. I've created #178 to address this and generally clean up that set of tests.

PR #85, opened 2022-05-07
Coerce booleans when constructing options object

cjbarth:
# Description

Improve handling of booleans by coercing incoming values when building the options object. This will simplify reasoning on variables down-the-line, potentially increase security (see https://github.com/node-saml/node-saml/pull/83#discussion_r866304349), and also allow for more JS JIT optimizations in the case where `"true"` was passed instead of `true`.


markstos:
This looks good to squash and merge. 

I presume the reason these inputs aren't typed as Boolean is for backcompat?

Or is a TypeScript translation issue were the type checks for run time get lost in translation?

An alternative to this approach then is a breaking change release which implements the cleaner change of typing these config values as booleans and letting the type system report the errors. But perhaps that's not even possible due to how TypeScript works.

cjbarth:
The reason for this has nothing to do with the type system. These are already typed at Boolean, and they even have defaults. The problem is when working with non-TypeScript users. If they pass in something like `"false"` instead of `false`, it will coerce to `true`, and the type checking system is always compiled out of TypeScript. So, we'd like to force that coercion early on in the code, in the `ctor`, so that down the line we don't have to worry about the type of comparison operators that we use, e.g. `==` vs `===`.

One possible option would be to actually have an assert that will check that the incoming values are genuine Boolean values and throw if they are not. In fact, that might be the better option so that there is no coercion that takes place, even in the `ctor`. I'll push up a change that does that.

cjbarth:
Actually @markstos , do you see value in a simple dependency like https://github.com/elierotenberg/typed-assert in order to assert the incoming types so that we support non-TypeScript users as first-class users? Right now there is nothing that strictly validates incoming data to make sure it complies with our requirements if TypeScript isn't being used. Otherwise, I feel like we'd end up implementing this ourselves.

I also looked at the similar [Zod](https://github.com/colinhacks/zod) library, but that seems more complex than we really need.

Please let me know what you think.

cjbarth:
@markstos Do you have any opinions on [this comment](https://github.com/node-saml/node-saml/pull/85#issuecomment-1124449318)? I'd like to see if that can land freeing up my headspace to work on some of these other PRs. I'd like to get as many of them landed as possible before the big 4.0 release.

Now that we have an org, I can do beta releases after every landed PR :)

cjbarth:
I've decided to just roll my own assertion where it made sense instead of taking on another dependency.

PR #90, opened 2022-06-16
Make Issuer Required in the Types Too (like it is at runtime)

### LINKED ISSUE #89: [BUG] Issuer is Now Required at Runtime But Not During Type Checking
**To Reproduce**

I adjusted the "get Urls" test from the samlTests.spec.ts file to omit the issuer (mimicking our app's previous setup), and got a runtime error of "issuer is required" when it was run
```ts
diff --git a/test/samlTests.spec.ts b/test/samlTests.spec.ts
index 815691b..0fed2df 100644
--- a/test/samlTests.spec.ts
+++ b/test/samlTests.spec.ts
@@ -17,7 +17,7 @@ describe("SAML.js", function () {
         entryPoint: "https://exampleidp.com/path?key=value",
         logoutUrl: "https://exampleidp.com/path?key=value",
         cert: FAKE_CERT,
-        issuer: "onesaml_login",
+        // issuer: "onesaml_login",
       });
       req = {
         protocol: "https",
```
```
1) SAML.js
       get Urls
         "before each" hook for "calls callback with original query string":
     TypeError: issuer is required
      at assertRequired (src/utility.ts:2:667)
      at SAML.initialize (src/saml.ts:2:4711)
      at new SAML (src/saml.ts:2:4242)
      at Context.<anonymous> (test/samlTests.spec.ts:16:14)
      at processImmediate (internal/timers.js:464:21)
```

**Expected behavior**
The TSC should've have failed beforehand when checking types instead

**Environment**

- Node.js version: v14.19.0
- node-saml version: Tip of master branch - commit SHA: abfb20d668b148f63e4acf8892f87853e3a9c168

sadikinfosec:
# Description

Resolves #89 

We had an application using the [4.0.0-beta.2](https://www.npmjs.com/package/node-saml/v/4.0.0-beta.2) version that we needed to update to use the latest master commit instead to pick up the update to the xml-encryption dependency that patched a security vulnerability. 

That app was not previously passing a value for "issuer" to the SAML constructor, but was relying on the library to provide a default.

After updating, this caused problems similar to the ones mentioned in the Github issue, and after rolling back and trying again, we patched this by providing a value for "issuer" to fit the new requirement.

We noted this could've been caught earlier had TS warned us about the new requirement, and so wanted to try and update the library to that effect for anyone else who might come across this change in the future.

# Checklist:

- Issue Addressed: 89 
- Link to SAML spec: N/A
- Tests included? Updated existing ones
- Documentation updated? 
    - The READme only seems to have examples where issuer is set. I'm not sure exactly where this would make sense to mention

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -50,6 +50,7 @@ export type CertCallback = (
  */
 export interface MandatorySamlOptions {
   cert: string | string[] | CertCallback;
+  issuer: string;
 }
 
 export interface SamlIDPListConfig {
@@ -102,7 +103,6 @@ export interface SamlOptions extends Partial<SamlSigningOptions>, MandatorySamlO
   protocol?: string;
   host: string;
   entryPoint?: string;
-  issuer: string;
   decryptionPvk?: string | Buffer;
 
   // Additional SAML behaviors
--- a/test/crypto.spec.ts
+++ b/test/crypto.spec.ts
@@ -4,7 +4,6 @@ import * as assert from "assert";
 import { certToPEM, generateUniqueId, keyToPEM } from "../src/crypto";
 import { TEST_CERT } from "./types";
 import { assertRequired } from "../src/utility";
-import { SamlConfig } from "../src/types";
 
 describe("crypto.ts", function () {
   describe("keyToPEM", function () {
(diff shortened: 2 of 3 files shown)


cjbarth (approved):
Thanks for catching this.

cjbarth:
@sadikinfosec , could you please have a look at https://github.com/node-saml/node-saml/issues/246 and comment on how your fix appears to be breaking his case?

PR #91, opened 2022-06-24
Move to NPM organization

cjbarth:
# Description

This will move this package to the NPM @node-saml organization. After this, we'll publish this as a 4.0 beta and then update all the badges in the README.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "node-saml",
+  "name": "@node-saml/node-saml",
   "version": "3.0.0",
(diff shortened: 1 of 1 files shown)


markstos (approved):
This matches docs I've seen.

It seems NPM will not redirect the old requests to the new repo, though. We'll have to deprecate the old name space and put a notice there, and then all users will manually update their package.json references it seems. 

https://stackoverflow.com/questions/43007471/how-to-migrate-npm-package-to-an-organization-scope

PR #116, opened 2022-07-09
using 'protected' instead of 'private'

random42:
Hi,

I just changed all the private methods to protected. This allows for implementing extensions of the SAML protocol (like [SPID](https://developers.italia.it/it/spid/) which I'm working on) in an easy way.

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -116,7 +116,7 @@ async function promiseWithNameID(nameid: Node): Promise<NameID> {
 }
 
 class SAML {
-  // note that some methods in SAML are not yet marked as private as they are used in testing.
+  // note that some methods in SAML are not yet marked as protected as they are used in testing.
   // those methods start with an underscore, e.g. _generateLogoutRequest
   options: SamlOptions;
   // This is only for testing
@@ -191,7 +191,7 @@ class SAML {
     return options;
   }
 
-  private getCallbackUrl(host?: string | undefined) {
+  protected getCallbackUrl(host?: string | undefined) {
     // Post-auth destination
     if (this.options.callbackUrl) {
       return this.options.callbackUrl;
@@ -210,7 +210,7 @@ class SAML {
     }
   }
 
-  private signRequest(samlMessage: querystring.ParsedUrlQueryInput): void {
+  protected signRequest(samlMessage: querystring.ParsedUrlQueryInput): void {
     assertRequired(this.options.privateKey, "privateKey is required");
 
     const samlMessageToSign: querystring.ParsedUrlQueryInput = {};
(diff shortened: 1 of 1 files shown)


cjbarth:
Do you have a link to SPID that is in English?

random42:
[This](https://developers.italia.it/en/spid/) is the english page. SPID is the public authentication service for italian citizens.

random42:
The documentation is in italian, but for example the [AuthnRequest](https://docs.italia.it/italia/spid/spid-regole-tecniche/it/stabile/single-sign-on.html#authnrequest) needs to be changed a bit.

The `saml:Issuer` element must have attributes `Format` set to `urn:oasis:names:tc:SAML:2.0:nameid-format:entity` and `NameQualifier` set to `${issuer}`. They also want the `samlp:NameIDPolicy`'s `AllowCreate` attribute to disappear (false is not enough). This was the reason for me to have protected methods.

They also have many other elements needed for the metadata xml, but since that method is public that was not a problem for me.

IMHO, I don't see the big deal here. Having protected methods just allows for better extendibility.

cjbarth:
I'm not trying to make a big deal, just trying to understand. We can make them all protected, that suits me just fine, however, I'd like for the changes needed to be added directly to the project instead of to a subclass. If they were all added directly to the project, I'm not sure why `private` vs `protected` would even be a discussion.

Are the needed options something that can be added directly to the project, or are they contradictory with the SAML spec in some way?

random42:
Seems like the `Format` attribute of the `Issuer` can be part of SAML (see [here](https://en.wikipedia.org/wiki/SAML_2.0#SAML_attribute_query)) but I found `NameQualifier` only in SPID.
The `AllowCreate` attribute I think can be omitted instead of setting it to `false` as default behaviour, it should not break the SAML protocol. But honestly I don't know so much about it.

random42:
That's the case with format `urn:oasis:names:tc:SAML:2.0:nameid-
format:persistent`, but not `urn:oasis:names:tc:SAML:2.0:nameid-format:entity` that is the SPID format.

cjbarth:
`persistent` requires certain things, but other options are valid, and have different requirements. See https://stackoverflow.com/a/51581573/271351 and https://docs.oracle.com/cd/E19316-01/820-3886/ggwbz/index.html

We already support the config property `identifierFormat` and `spNameQualifier`, so it appears that we only need to have `nameQualifier` added, and then you can add to the config as needed and things should just work.

random42:
Sure, we can do it in another PR

cjbarth:
We can close this PR because #129 covers these changes and we are waiting for another PR from @random42  to add the SAML features that are needed.

PR #129, opened 2022-07-22
Refactor generate functions to a separate file

cjbarth:
In the spirit of #19, move some functions out of the very long `saml.ts` to another file without breaking class semantics. This will make it easier to land PRs with less chance of conflicts going forward.

### DIFF
--- a/src/metadata.ts
+++ b/src/metadata.ts
@@ -1,12 +1,12 @@
-import { removeCertPEMHeaderAndFooter } from "../crypto";
+import { removeCertPEMHeaderAndFooter } from "./crypto";
 import {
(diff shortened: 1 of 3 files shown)


markstos:
If I'm reading this correctly, one key difference with #19, is that here the smaller file that's broken out still retains a `this` that refers to a SAML object, while #19 took a more aggressive approach by creating functions that have explicit inputs and outputs.

I'm going to follow-up more in #131. 

So while #19 prioritizes code that is easier to reason about, this PR takes a  compromise approach of splitting up files which might help avoid conflicts. This PR also retains more internal API compatibility.

PR #130, opened 2022-07-23
Refactor process routines out of saml.ts

cjbarth:
In the spirit of #19, move some functions out of the very long `saml.ts` to another file without breaking class semantics. This will make it easier to land PRs with less chance of conflicts going forward.

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -17,17 +17,14 @@ import {
   Profile,
   SamlOptions,
   SamlConfig,
-  XMLObject,
   XMLOutput,
   ValidateInResponseTo,
-  XMLValue,
 } from "./types";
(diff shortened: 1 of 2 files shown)


cjbarth:
@markstos I would like your honest opinion on these PRs (#129, #130) before I continue. Do you think this adds to, or detracts from, the maintainability, debug-ability, comprehensibility, or any other aspect of the code by breaking it up like this?

I note that in VS Code "Find All References..." no longer works as one might expect when the code is refactored into multiple files. I'm not sure if losing that is worth any other gains we might get. After seeing how this looks so far, I'm starting to see how this might not be the 100% win it seemed it might be.

@srd90, @zoellner, @sibelius, @josecolella, @gugu since you also spend a lot of time with this code, I'd value your opinion as well.

PR #140, opened 2022-08-02
Add support for metadata ContactPerson and Organization

Its-Just-Nans:
# Description

Hello,
What do you think about this ?

- Add the possibility to add more fields in the metadata
- Update `package.json` to support the type `Buffer`
- Add type
- Add test

# Checklist:

- [x] Link to SAML spec: [https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf) it's quite general :
   - `Element <Organization>` (page 12)
   - `Element <ContactPerson>` (page 12)
   - `AdditionalMetadataLocation>` (page 14)
- [x] Tests included
- [x] Documentation updated (**in `README.md`**)


test/static/expected_metadata_metadataExtensions.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0"?>
+<EntityDescriptor xmlns="urn:oasis:names:tc:SAML:2.0:metadata" xmlns:ds="http://www.w3.org/2000/09/xmldsig#" entityID="http://example.serviceprovider.com" ID="d700077e-60ad-49c1-b93a-dd1753528708">
```
cjbarth, line 2: If you look at page 6 of the spec, it seems you might be missing a namespace declaration for `md:*` properties: e.g. `xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata`.

Its-Just-Nans: You're are right.
I fixed it

package.json
```diff
@@ -69,7 +69,7 @@
     "@istanbuljs/nyc-config-typescript": "^1.0.2",
     "@types/chai": "^4.3.1",
     "@types/mocha": "^9.1.1",
-    "@types/node": "^14.18.22",
+    "@types/node": "^14.18.23",
```
cjbarth, line 72: I see that there are merge conflicts. If you back out this update and the `package-lock.json` updates, the conflicts should go away. We update dependencies regularly, so you don't need to worry about that.

Its-Just-Nans: Okay I changed, it's because I had an typescript error with `Buffer` type

cjbarth: So, with this change, are you now having problems with the typing? If so, we can make a PR to update those types for you.

Its-Just-Nans: Hum the type looks fine, I think my VScode was buggy and didn't detect the `Buffer` type

test/static/expected_metadata_metadataExtensions.xml
```diff
@@ -0,0 +1,48 @@
+<?xml version="1.0"?>
+<EntityDescriptor xmlns="urn:oasis:names:tc:SAML:2.0:metadata" xmlns:ds="http://www.w3.org/2000/09/xmldsig#" xmlns:md="urn:oasis:names:tc:SAML:2.0:metadata" entityID="http://example.serviceprovider.com" ID="d700077e-60ad-49c1-b93a-dd1753528708">
```
cjbarth, line 2: Ok, I might have mislead because I read what you were doing too quickly; now you have the same namespace listed twice here. The problem is that what you are adding isn't a metadata _extension_, it is just metadata. If you look at the spec on line 385 you'll see that a non-SAML namespace should be used for extensions.

What you're trying to do is identified on lines 432-434 and is a well-known SAML spec, so we should provide some structure for it instead of allowing just any object to be inserted into the metadata (following our existing pattern). Each of the three types of metadata you mention in the description of this PR should have its own set of properties, or take an object, or probably an array of objects, that it duck-typed at run-time and ideally type-checked by TypeScript.

Its-Just-Nans: I've made a fix

src/types.ts
```diff
@@ -153,6 +153,31 @@ export interface SamlOptions extends Partial<SamlSigningOptions>, MandatorySamlO
   disableRequestAcsUrl: boolean;
   samlAuthnRequestExtensions?: Record<string, unknown>;
   samlLogoutRequestExtensions?: Record<string, unknown>;
+  metadataExtensions?: {
```
cjbarth, line 156: This is a misleading name. These aren't metadata extensions, they are metadata elements (line 443). I think this should be broken up into individual configuration values, like `metadataContactPerson` and `metadataOrganization`. Otherwise, I think this looks good.

I take it that for your use, you only need `ContactPerson` and `Organization`? That is totally fine with me; only implement the part of the spec you need. As long as they are spec compliant and work for you we're good :)

Its-Just-Nans: Done

cjbarth (approved):

PR #141, opened 2022-08-02
Improve tests

Its-Just-Nans:
# Description

- Transform `url.parse()` into `new URL()`
- Add eslint deprecation plugins
- Add one test for `generateServiceProviderMetadata throw error`

# Checklist:

- [x] Tests included

### DIFF
--- a/test/tests.spec.ts
+++ b/test/tests.spec.ts
@@ -1,6 +1,6 @@
 "use strict";
 import { SAML } from "../src/saml";
-import * as url from "url";
+import { URL } from "url";
 import * as querystring from "querystring";
 import { parseString, parseStringPromise } from "xml2js";
 import * as fs from "fs";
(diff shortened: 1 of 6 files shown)


test/tests.spec.ts
```diff
@@ -571,6 +571,31 @@ describe("node-saml /", function () {
...
+          // typescript
+          if (error instanceof Error) {
+            expect(error.toString()).to.contain(
+              "Error: Missing decryptionCert while generating metadata for decrypting service provider"
+            );
+          }
+        }
+        expect(metadata).to.be.undefined;
+      });
+
```
markstos, line 598: Additional test coverage is welcome, but what does this have to with the change from `url.parse() to `new URL()`?

Its-Just-Nans: Its a legacy API

The Node.sj documentation said `Use the WHATWG URL API instead.`

- [https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost](https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost)

The possible argument tu use `url.parse` are :
- `Faster than the alternative WHATWG URL parser.` - it is used in `test` files so I think it's not the purpose here
- `Easier to use with regards to relative URLs than the alternative WHATWG URL API.` - the `new URL` works good in our case

Its-Just-Nans: But I agree it can be split into 2 PR

markstos:
For added context, the new `URL` class was added in Node 10.

In Node 11, `url,parse()` was deprecated, but in Node 14, the deprecation was reversed (!!) and marked as "Legacy" instead. 

Still, because `new URL()` is the recommended syntax and supported on all the versions of NOde we support, I recommend approving that part of the change.

I left a comment about the additional test coverage that I'm not sure relates.

test/tests.spec.ts
markstos: I'm not concerned about the performance difference in the URL parses. SAML would a small amount of traffic for most web-apps and URL parsing is probably not the bottleneck. 

I'm OK with leaving this as one PR as long as @cjbarth agrees about accepting the additional test coverage.

Its-Just-Nans: Okay great

I can also split the PR if needed 😃

cjbarth (approved):

PR #143, opened 2022-08-04
Include AuthnRequestsSigned attribute in all metadata

Its-Just-Nans:
# Description

Add `AuthnRequestsSigned="false"` if the options is defined

# Checklist:

- [x] Link to SAML spec: [https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf) page 18: optionnal param
- [x] Tests included
- [x] Documentation updated


README.md
```diff
@@ -60,6 +60,7 @@ const saml = new SAML(options);
 - `allowCreate`: grants permission to the identity provider to create a new subject identifier (default: `true`)
 - `spNameQualifier`: optionally specifies that the assertion subject's identifier be returned (or created) in the namespace of another service provider, or in the namespace of an affiliation of service providers
 - `wantAssertionsSigned`: if truthy, add `WantAssertionsSigned="true"` to the metadata, to specify that the IdP should always sign the assertions.
+- `authnRequestsSigned`: if defined, add `AuthnRequestsSigned="false"` to the metadata, to specify that the request is not signed. This option is canceled if you sign your requests.
```
cjbarth, line 63: Do I understand correctly that if this value is set to `true` or `false`, then `AuthnRequestsSigned="false"` will be added to the metadata? Why use a boolean that way? Also, what do you mean by an option being canceled?

Its-Just-Nans: The option will be overrided here is it's a signed request.
[https://github.com/node-saml/node-saml/blob/a7de497ba8008cf88f1bc8d5c212e5d268254be3/src/metadata.ts#L69](https://github.com/node-saml/node-saml/blob/a7de497ba8008cf88f1bc8d5c212e5d268254be3/src/metadata.ts#L69)

This options is used in a "logical" way

If a use it a a boolean two use case can appear:
- putting `false` while request is signed ? -> it will be overrided
- putting `true` if the request is not signed -> it is not correct

test/tests.spec.ts
```diff
@@ -588,6 +588,21 @@ describe("node-saml /", function () {
         expect(metadata).to.contain('AuthnRequestsSigned="true"');
       });
 
+      it("generateServiceProviderMetadata contains AuthnRequestsSigned=false", function () {
+        const samlConfig: SamlConfig = {
+          cert: TEST_CERT,
+          issuer: "http://example.serviceprovider.com",
+          callbackUrl: "http://example.serviceprovider.com/saml/callback",
+          identifierFormat: "urn:oasis:names:tc:SAML:2.0:nameid-format:transient",
+          authnRequestsSigned: false,
```
cjbarth, line 597: When I pull this code and change `authnRequestsSigned: false` to `authnRequestsSigned: true`, the test still passes, so I must be missing something about the way you intend this to work.

Its-Just-Nans: Yes I explained it here: 
https://github.com/node-saml/node-saml/pull/143#discussion_r938196687

This options is used in a "logical" way

If we use the option as a boolean, two weird case can appear:

- if we put `authnRequestsSigned: false` while request is signed ? -> it will be overridden, because in the code, we force `authnRequestsSigned` to be true
- if we put `authnRequestsSigned: true` but the request is not signed (for example a missing cert) -> the SAML metadata will not be correct, because the request isn't signed

To solve that, the option is: "if defined, put the options" (the explanation is in the `README.md`)

cjbarth: I see what you're trying to do now. According to the [spec](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf) line 689, this should default to false. So, why not just include it all the time and let it get overridden when it should be true as it currently does? Put another way, what value is added by having the ability to _not_ include `AuthnRequestsSigned="false"`?

Its-Just-Nans: Note that the change is for `AuthnRequestsSigned` line 740 in [spec](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf)

> why not just include it all the time and let it get overridden when it should be true as it currently does?

- If, for example, you want a smal metadata xml file, you don't want all an optional field
- If we include it all the time, I think we need to rewrite tests :/

You choose, I can implement the `include by default` if you want

cjbarth: Yea, I like the idea that it is included by default. It doesn't add that much to the metadata and it makes it very clear what is going on. I've pushed up making this the default, but it needs some cleanup to remove the option, readme changes, etc. Would you please follow?

Its-Just-Nans: done !

cjbarth (approved):

PR #151, opened 2022-08-16
Remove insecure clockSkew recommendation

cjbarth:
Based on [this discussion](https://github.com/node-saml/node-saml/pull/83#issuecomment-1119082248), it seems we are needlessly recommending an insecure configuration and such a recommendation is being copied by others. This removes the insecure recommendation.


markstos:
The ADFS document was very old-- it pre-dates my involvement and I am not familiar with ADFS personally. I'm fine with removing this. If someone who uses passport-saml and ADFS wants to maintain their own documentation, they can. This doesn't need to be in the core project repo.

markstos (approved):

PR #152, opened 2022-08-16
Reduce frequency of dependabot updates

cjbarth:
Dependabot can get a little noisy with its notifications about updates. Reduce the frequency while still retaining information about everything that needs updating and the convenient PR that is ready-to-go.

See https://github.com/node-saml/node-saml/pull/150#issuecomment-1216795167


markstos:
Thanks!

PR #157, opened 2022-08-18
feat: add metadataID

Its-Just-Nans:
# Description

As discussed here [https://github.com/node-saml/node-saml/discussions/156](https://github.com/node-saml/node-saml/discussions/156)

# Checklist:

- [x] Link to SAML spec: [https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf) line 374
- [x] Tests included
- [x] Documentation updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,8 @@ const saml = new SAML(options);
 - `protocol`: protocol for callback; will be combined with path and server host information to construct callback url if `callbackUrl` is not specified (default: `http://`)
 - `host`: host for callback; will be combined with path and protocol to construct callback url if `callbackUrl` is not specified (default: `localhost`)
 - `entryPoint`: identity provider entrypoint (is required to be spec-compliant when the request is signed)
-- `issuer`: issuer string to supply to identity provider
-- `metadataID`: (optional) string to supply for the field `ID` in the generated xml metadata file from `generateServiceProviderMetadata` function
+- `issuer`: issuer string to supply to identity provider (for the `entityID` field)
+- `metadataID`: (optional) string to supply for the field `ID` of `EntityDescriptor` in the generated xml metadata file from the `generateServiceProviderMetadata` function
 - `audience`: expected saml response Audience, defaults to value of Issuer (if `false`, Audience won't be verified)
 - `cert`: the IDP's public signing certificate used to validate the signatures of the incoming SAML Responses, see [Security and signatures](#security-and-signatures)
 - `privateKey`: see [Security and signatures](#security-and-signatures).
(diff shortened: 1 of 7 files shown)


barryhagan:
A unique default value for the metadata ID was introduced in #30 but was later regressed when code was shuffled around between files.  You should restore the use of `generateUniqueId` as the default behavior of the ID.

cjbarth:
> regressed when code was shuffled around between files

I'm glad you pointed this out @barryhagan . I'll have to take a look at where it regressed (would you happen to know the commit?). Feel free to recreate the PR to put it back. I'm going to try to do another beta release this weekend since there has been a lot of changes. I need to get the changelog updated before I do that though.

cjbarth:
Or, even better, would be to get all of @barryhagan requirements met in this PR as well and we can land them all at the same time :)

barryhagan:
https://github.com/node-saml/node-saml/commit/3d61e918b272a6bf32872b7b266c1244f9c4a87a#diff-4a2bc5d4a8380472458372e241f56bcaebbd02564452e03486e1dc0ea7120de3R63 is the regression commit.

It should be pretty easy to get that fixed as part of this PR, but the use case for setting a static custom ID for metadata is not clear to me.

cjbarth:
> regression commit

Thanks for that. I'll go over that carefully to see how that happened... and how come no tests broken (which is the biggest question in my mind).

barryhagan:
I put up a separate PR to fix the regression: https://github.com/node-saml/node-saml/pull/158

If you merge that first, you'll have to rebase or resolve this PR.

The tests didn't break because they used a replacement value [here](https://github.com/node-saml/node-saml/blob/de66a845da47f4759b4b36acac2dddddf2a73395/test/tests.spec.ts#L431) so the actual ID value was never checked, just rewritten.   This can now be fixed given that `generateUniqueId` was later changed to a configurable function.  I updated the tests to do this in the PR.

barryhagan:
>Regarding random EntityDescriptor ID : IMHO SAML core spec snippet at https://github.com/node-saml/node-saml/pull/30 regarding additional requirements on top of those of xs:ID are for assertions, requests and responses (not for SP/IdP metadata's XML element IDs).

I'd say that is more ambiguity in the spec than intentional omission of referencing metadata.

What this boils down to is the `ID` is used as a reference for signing documents so it really should be opaque and transient, not static.  When you turn on the `signMetadata` option, the signature's `Reference` is the `ID` in the metadata EntityDescriptor.   I think the spec is trying to guard against signing validation bugs.  (What if you already saw and validated a section with the same ID before? Maybe you don't revalidate it and you have a security hole)

That is why #30 changes the metadata ID to be unique each time it is generated and I am skeptical of allowing it to be statically configured.

Its-Just-Nans:
In the end, we can close this one

PR #177, opened 2022-09-20
Require all assertions be signed; add option to disable

cjbarth:
# Description

As a follow-on to #83, require all assertions to be signed and provide the option to defeat this requirement.


cjbarth:
I appreciate your comments about the "old" behavior, and I've added tests to cover those cases.

`root-invalidly-signed-assert-invalidly-signed` and `root-modified-after-valid-signature-assert-modified-after-valid-signature` are not conditions that we currently test for, but it seems we generally cover those cases. Invalid signature code paths are already tested, and are the same as unsigned code paths (though there is an early exit in that case). Modifying signed content after the signature makes the signature invalid, so the above still applies.

Otherwise, I can't see any specific cases where we are failing to thoroughly exercise the code. If so, we'll do another PR to add to the existing test suite. This PR is trying to just adjust the default.

PR #184, opened 2022-10-06
test: add test coverage for invalid XML with multiple root nodes.

markstos:


markstos:
My test logic was backwards here. Converting back to draft.

markstos:
To replaced by another approach.

PR #190, opened 2022-10-14
Update documentation to remove ADFS references; rename passport-saml

cjbarth:
# Description

In preparation for a 4.x release, we want to get the documentation cleaned up. The documentation for this library should be focused on configuration of the library. The IdP-specific directions should be either at `passport-saml` or in a separate file in this project and shouldn't reference Passport at all.

References #151


markstos (approved):
This is good. We don't have the bandwidth to maintain IdP-specific docs. Community-contributed "Discussions" or "Wiki" are better suited for that.

PR #194, opened 2022-10-20
feat: expose getLogoutResponseUrlAsync publicly

### LINKED ISSUE #193: [ENHANCE] getLogoutResponseUrlAsync should be exposed
I see that getLogoutResponseUrl is just a wrapper around getLogoutResponseUrlAsync, so my question is simply why is that last method not public? This forces users to re-wrap that call into a promise when used in an async/await context.

Change from protected to public.

I'm not sure there's a need for alternatives.

Marsup:
# Description

Expose getLogoutResponseUrlAsync publicly.
Fixes #193.
I have duplicated the tests from the callback version with the necessary changes for async, I'm not sure whether you want to factor those tests.

# Checklist:

- Issue Addressed: [X]
- Link to SAML spec: [ ]
- Tests included? [X]
- Documentation updated? [ ]

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -620,7 +620,7 @@ class SAML {
     )(callback);
   }
 
-  protected async getLogoutResponseUrlAsync(
+  async getLogoutResponseUrlAsync(
     samlLogoutRequest: Profile,
     RelayState: string,
     options: AuthenticateOptions & AuthorizeOptions,
--- a/test/samlTests.spec.ts
+++ b/test/samlTests.spec.ts
@@ -1,4 +1,5 @@
 "use strict";
+import * as sinon from "sinon";
 import { URL } from "url";
 import { expect } from "chai";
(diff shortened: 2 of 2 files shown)


cjbarth:
A quick look over this code from my phone seems to indicate you just copied the test and adjust them. I don't know that we need duplicate tests. I think what we really need to test is making sure that the public contract is correct.

Perhaps that could be a simple as making a call to the async function and then making a call to the callback function and making sure the results are the same except the async function should have returned it in a promise.

Marsup:
Unfortunately, it doesn't work, SAMLResponse are different, I've tried with fake timers but still different, any idea what could be wrong?

Marsup:
OK, I found out it was about the unique id generator, it should be fine now.

cjbarth (approved):

cjbarth (approved):

PR #195, opened 2022-10-23
Throw if multiple XML roots detected

cjbarth:
# Description

Based on the discussion at #191, it seems there is a better way to protect against XML files with multiple roots. This PR is to implement that.

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -677,6 +677,14 @@ class SAML {
       if (!Object.prototype.hasOwnProperty.call(doc, "documentElement"))
         throw new Error("SAMLResponse is not valid base64-encoded XML");
 
+      if (
+        Array.from(doc.childNodes as NodeListOf<Element>).filter(
+          (n) => n.tagName != null && n.childNodes != null
+        ).length !== 1
+      ) {
+        throw new Error("Malformed XML; multiple roots detected");
+      }
+
       const inResponseToNodes = xpath.selectAttributes(
         doc,
         "/*[local-name()='Response']/@InResponseTo"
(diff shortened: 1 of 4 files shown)


cjbarth:
I've made some changes to take into considerations @srd90's latest comments about how other parts of the code could use the same XML validation.

markstos (approved):
I agree that throwing if we detect multiple SAML roots is a stronger approach than our first attempt.

I saw that `parseDomFromString` switched to an async method that returned a promise. I reviewed the docs to see if this was a public method and thus a breaking change, but it appears that it was not a public method and this not a breaking change.

frumioj:
I think this is the best place for you to make this specific change, and it looks correct to me at least reading the code. Thanks for doing this. 

The only thing which could be stronger would be to not even parse as far as the end of the AuthnResponse, and return an authentication error when the status is read (earlier than the end of the response) but I'm guessing this would be harder because the parse strategy would have to change (SAX interface vs DOM). 

I will try to look at whether that is even possible or a good idea.

The other thing I'll note is that I submitted a similar comment to the security list for the `xmldom` package, and they may change their behaviour to error/throw when multiple root nodes are discovered. I don't think that will change this approach, but it might make it unnecessary.

frumioj:
> The other thing I'll note is that I submitted a similar comment to the security list for the `xmldom` package, and they may change their behaviour to error/throw when multiple root nodes are discovered. I don't think that will change this approach, but it might make it unnecessary.

https://github.com/xmldom/xmldom/security/advisories/GHSA-crh6-fp67-6883

PR #209, opened 2022-11-09
Fix #207, updated @xmldom/xmldom and xml-crypto packages

### LINKED ISSUE #207: [SECURITY]: Fix @xmldom/xmldom audit issue GHSA-crh6-fp67-6883
`@xmldom/xmldom` library has critical vulnerability https://github.com/advisories/GHSA-crh6-fp67-6883

Update @xmldom/xmldom to 0.8.5

RopoMen:
Updated
- @xmldom/xmldom
- xml-crypto

Fixed unit test, because now xmldom detects multiple roots.

### DIFF
--- a/src/xml.ts
+++ b/src/xml.ts
@@ -189,7 +189,6 @@ export const parseDomFromString = (xml: string): Promise<Document> => {
     function errHandler(msg: string) {
       return reject(new Error(msg));
     }
-
     const dom = new xmldom.DOMParser({
       /**
        * locator is always need for error position info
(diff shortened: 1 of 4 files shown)


RopoMen:
This fixes also PR https://github.com/node-saml/node-saml/pull/205/

cjbarth:
Closed in favor of https://github.com/node-saml/node-saml/pull/213 and https://github.com/node-saml/node-saml/pull/205

src/xml.ts
```diff
@@ -209,14 +208,6 @@ export const parseDomFromString = (xml: string): Promise<Document> => {
...
     }
 
-    if (
-      Array.from(dom.childNodes as NodeListOf<Element>).filter(
-        (n) => n.tagName != null && n.childNodes != null
-      ).length !== 1
-    ) {
-      return reject(new Error("Malformed XML; multiple roots detected"));
-    }
-
```
RopoMen, line 219: @cjbarth I noticed that you already merged another PR which updated xmldom. That is ok, but I noticed that there was added comment https://github.com/node-saml/node-saml/blob/master/src/xml.ts#L217

My opinion is that that is dead code that should be removed.

cjbarth: I was thinking about it, then I realized that this code actually protects against a vulnerability and was torn about removing it and then potentially a dependency reintroduces the problem and we are vulnerable again. However, as I type that, I realize they have put in tests to make sure that doesn't happen, so we probably should just remove the old code. I'll create another PR for that.

PR #215, opened 2022-11-15
Remove check now covered by dependency

cjbarth:
# Description

After a security fix was made in `@xmldom/xmldom@0.8.5`, some code is no longer needed in `node-saml`; removing.

### DIFF
--- a/src/xml.ts
+++ b/src/xml.ts
@@ -209,15 +209,6 @@ export const parseDomFromString = (xml: string): Promise<Document> => {
       return reject(new Error("Not a valid XML document"));
     }
 
-    if (
-      Array.from(dom.childNodes as NodeListOf<Element>).filter(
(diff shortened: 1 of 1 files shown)


RopoMen:
@cjbarth your comment in that closed PR 
> However, as I type that, I realize they have put in tests to make sure that doesn't happen...

Remember that `node-saml` has its own test for "multiple roots" to ensure that even if xmldom makes some stupid change, tests in this projects will catch it, see: https://github.com/node-saml/node-saml/blob/master/test/test-signatures.spec.ts#L72-L77

That test will fail if xmldom is not causing that error what test is expecting. Error is thrown over here https://github.com/node-saml/node-saml/blob/master/src/xml.ts#L190

(and it is best that node-saml is not "translating" those xmldom errors to our own versions)

PR #220, opened 2022-11-17
Expose ValidateInResponseTo as it is required in options

ghusse:
# Description

Options required by `passport-saml` when creating a new configuration require the property `validateInResponseTo` to use the enum `ValidateInResponseTo`. But this enum is not easily accessible in `@node-saml/passport-saml`.

In typescript, we need to import `@node-saml/node-saml/lib/types` to use this enum, which requires to add the dependency to `@node-saml/node-saml` to a project that would only require `@node-saml/passport-saml`.

By exposing this enum in `node-saml`, it'll be also exposed by `passport-saml` once updated to the latest version of its dependency.

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [ ]
- Tests included? [ ]
- Documentation updated? [ ]

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -6,6 +6,7 @@ import {
   Profile,
   SamlConfig,
   SamlOptions,
+  ValidateInResponseTo,
 } from "./types";
 
-export { SAML, CacheItem, CacheProvider, SamlOptions, MandatorySamlOptions, Profile, SamlConfig };
+export { SAML, CacheItem, CacheProvider, SamlOptions, MandatorySamlOptions, Profile, SamlConfig, ValidateInResponseTo };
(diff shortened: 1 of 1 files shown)


cjbarth (approved):

cjbarth:
@ghusse , even though I merged this, please see some of these other linked PRs and offer your comments. There seems to be a bit of discussion around types and I want to make sure your experience it taken into account.

cjbarth:
I will land this and do releases once we figure out what the best approach forward is. It does seem though that if we stop exporting everything, then it will have to wait to be a semver-major change. If we just add to and re-work the exports, we can keep it as a semver-minor change; we can always remove the extra exports with v5 Q1 2023.

PR #221, opened 2022-11-17
fix: correct handling of XML entities in signature attributes

Marsup:
# Description

Fixes #200.

XML entities, when included in signature elements (value and/or certificate) fail signature check, passing the XML node to xml-crypto instead of providing a string fixes the issue.

# Checklist:

- Issue Addressed: [X]
- Link to SAML spec: [ ]
- Tests included? [X]
- Documentation updated? [ ]

### DIFF
--- a/src/xml.ts
+++ b/src/xml.ts
@@ -129,8 +129,7 @@ export const validateXmlSignatureForCert = (
     getKeyInfo: () => "<X509Data></X509Data>",
     getKey: () => Buffer.from(certPem),
   };
-  const signatureStr = normalizeNewlines(signature.toString());
-  sig.loadSignature(signatureStr);
+  sig.loadSignature(signature);
   // We expect each signature to contain exactly one reference to the top level of the xml we
   //   are validating, so if we see anything else, reject.
   if (sig.references.length != 1) return false;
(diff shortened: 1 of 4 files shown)


Marsup:
We should be normalizing the signature (and only the signature) we provide through `loadSignature`, all paths that lead to this call appears to provide a node that comes out of `parseDomFromString`, so it is already normalized. This way we can keep `fullXml` the way it is. I guess another way to do it if you wanted to keep a string would be to normalize both `\r` and XML entities, but since you already have a normalized node, that seems like unnecessary work.

cjbarth:
Now that we have a test to make sure things continue working, we'll know if something else changes.

cjbarth (approved):

jonathansamines:
Hey @cjbarth, any chance we can get these changes published? I am working on moving some projects from `passport-saml` to `@node-saml/node-saml` and this is currently the only problem we have found so far. I was able to locally verify these changes and they work for our integrations ;)

cjbarth:
Released @jonathansamines .

jonathansamines:
awesome, thank you @cjbarth

grundb:
@cjbarth Great to see that you seem to have solved this problem! I noticed version 4.0.2 seems to be missing from npm, though: https://www.npmjs.com/package/@node-saml/passport-saml?activeTab=versions. Maybe you just haven't had time to publish it or something, but just letting you know :+1:

Marsup:
He doesn't need to, the semver range is perfectly appropriate.

PR #224, opened 2022-11-17
Export types required for SamlOptions

cjbarth:
# Description

There are several types referenced by SamlOptions which aren't exported for a consumer to use. This PR, in the vane of #220, corrects this problem.


cjbarth:
@repomen @ghusse, in continuation of our discussion [here](https://github.com/node-saml/passport-saml/pull/813), I'm seeking your eyes on this PR. I'm trying to get things ready for a v5 release and want to make sure all your typing concerns are met since we are in a semver-major breaking-change window with this update to a new version of Node.

ghusse:
I took a look at types, and I noted that some public methods from the SAML object were using types that are not exported.

I'm not sure that these methods are meant to be called by users of this lib (maybe they should not be public?), so I'll let you decide what to do for each of them:

- `getAuthorizeUrlAsync` uses `AuthorizeOptions`
- `getLogoutUrlAsync`, `getLogoutResponseUrl` and `getLogoutResponseUrlAsync` use `AuthenticateOptions` and `AuthorizeOptions`
- `validateRedirectAsync` uses `ParsedQs`

ghusse:
I checked the rest and it seems OK.

cjbarth:
@ghusse , thanks for checking things over. I think I've addressed your findings. Can you have one more look before I merge this?

ghusse:
Thanks for the changes. I still see that `validateRedirectAsync` is public and is requesting a `ParsedQs` which is exposed by a dependency but not by node-saml itself.

cjbarth:
True, but it is also available via other packages. So, I guess it should be a `peerDependency`. I hesitated to do that because it is only needed if the consumer is using TypeScript. Thoughts?

cjbarth:
Further examination reveals that `qs` comes from `express`, so we have a transient dependency on `express`. I'll make another PR to refactor to remove that dependency.

Issue #206, opened 2022-11-08
[ENHANCE]: add support for public keys


RopoMen:
Hi,

TL;DR; Allow Public Key usage through `cert` option.

## Background

I'm currently maintaining [suomifi-passport-saml Fork](https://github.com/vrk-kpa/suomifi-passport-saml) and we are trying to get back to upstream usage, because `node-saml` currently supports `SAML Authn Request Extensions`.

But node-saml is lacking support for plain public key usage, it only supports certs through `cert` configuration option. `suomifi-passport-saml` currently has this kind-of "hax" https://github.com/vrk-kpa/suomifi-passport-saml/blob/master/lib/passport-saml/saml.js#L546-L570 to allow public key usage, pretty trivial.

SAML Metadata can contain `KeyInfo` in two formats; `ds:RSAKeyValue` and/or `ds:X509Data` both must represent the same key. See http://docs.oasis-open.org/security/saml/Post2.0/sstc-metadata-iop.html
`2.5.1   Key Representation` 

> In the case of the latter, only a single certificate is permitted. If both forms are used, then they MUST represent the same key.

Also in same document, section `2.6.1   Key Processing`

> Each key expressed by a <md:KeyDescriptor> element within a particular role MUST be treated as valid when processing messages or assertions in the context of that role. Specifically, any signatures or transport communications (e.g., TLS/SSL sessions) verifiable with a signing key MUST be treated as valid, and any encryption keys found MAY be used to encrypt messages or assertions (or encryption keys) intended for the containing entity.

> ... a metadata consumer MUST extract the public key found in the certificate and MUST NOT honor, interpret, or make use of any of the information found in the certificate other than as an aid in identifying the key used...

This means that even if certificate is expired or it is found in revocation list, public key cannot be ignored, it must be treated as valid. x509 certificate can be seen as a container for the public key.

SAML idP's / SP's may be able to provide both key presentations or only `ds:RSAKeyValue` or only `ds:X509Data`. Usually there is some technical limitation which is not in "our hands" if only one of the key presentations is available.

## What should be done?

[...]

PR #225, opened 2022-11-18
Add public key support

RopoMen:
# Description
- re-factored "crypto.ts" to handle PEM files in more general way.
- Added unit tests for new functions.
- Re-factored internally implementation to use "key info" (or similar) terminology instead of talking only about certificates. Certificate term is ok, if we known that we are handling certificate.

These changes are not making any changes into used interface, only internal re-factoring.

I will make changes to README.md before merge, because it is not currently talking about public key support. Public key support comes through `cert` property, if PEM file is well formatted then it is accepted.


RopoMen:
Resolves https://github.com/node-saml/node-saml/issues/206

cjbarth:
Thank you for this, but all these code scanning alerts will have to be addressed. If you need help figuring out how to make the regex less catastrophic, please post here and we'll see what help we can provide.

It looks like the new code you added isn't 100% covered by tests. Please have a look and see what can be done about that.

I look forward to seeing your README updates; our README definitely needs a little TLC.

Also, if it is easy to add a test to add coverage to another part of the code that is currently untested, perhaps a branch, please consider that; sometimes I find that when I'm writing a test trying to get complete coverage for my new code, I see that I can make one tweak to a new test and cover a previously untested branch in some other code.

RopoMen:
@cjbarth 

1. `..but all these code scanning alerts will have to be addressed.` ofc. I'll check what is the best way to test potential fix before pushing changes. (I have not used Github code scanning tools before)
2. `It looks like the new code you added isn't 100% covered by tests. `, I'll check what improvements there can be made. Those PEM handling functions should be covered as much as possible. My guess is that assertions should be tested better.
3. README.md updates
4. Add at least one unit test for saml.ts where well formatted PEM public key is given into library.
5. Check that is there need to add tests for saml.ts `getKeyInfosAsPem()` function. (most likely yes)

cjbarth:
> 1. `..but all these code scanning alerts will have to be addressed.` ofc. I'll check what is the best way to test potential fix before pushing changes. (I have not used Github code scanning tools before)

One of the best ways is to just look at the manual: https://codeql.github.com/codeql-query-help/javascript/js-polynomial-redos/. As suggested, anchoring the string or using some lookbehind helps a lot.

RopoMen:
@cjbarth I have now;
- Added more tests for crypto.ts
- Fixed regex issues
- Added test for public key
- Added tests for resolveAndParseKeyInfosToPem()
- Updated readme to contain mention about public key support

src/crypto.ts
```diff
@@ -1,48 +1,93 @@
...
-export const removeCertPEMHeaderAndFooter = (certificate: string): string => {
-  // These headers and footers are standard: https://www.ssl.com/guide/pem-der-crt-and-cer-x-509-encodings-and-conversions/#ftoc-heading-1
-  certificate = certificate.replace(/-----BEGIN CERTIFICATE-----\r?\n?/, "");
-  certificate = certificate.replace(/-----END CERTIFICATE-----\r?\n?/, "");
-  certificate = certificate.replace(/\r\n/g, "\n");
-  return certificate;
+export const stripPemHeaderAndFooter = (certificate: string): string => {
+  return certificate
+    .replace(/(\r\n|\r)/g, "\n")
+    .replace(/-----BEGIN [A-Z\x20]{1,48}-----\n?/, "")
```
cjbarth, line 91: Why do you use `\x20` instead of a space?

src/saml.ts
```diff
@@ -40,14 +40,38 @@ import {
...
+}: Pick<SamlOptions, "cert">): Promise<string[]> => {
+  const keyInfosToHandle: Array<string | string[]> = [];
+  const pemFiles: string[] = [];
+  if (typeof cert === "function") {
+    await util
+      .promisify(cert as CertCallback)()
+      .then((certs) => {
+        assertRequired(certs, "callback didn't return cert");
+
+        keyInfosToHandle.push(certs);
```
cjbarth, line 62: You might have to use a spread operator here as we don't want to have nested arrays.

test/crypto.spec.ts
```diff
@@ -43,22 +96,144 @@ describe("crypto.ts", function () {
...
-      }
+  describe("keyInfoToPem", function () {
+    describe("invalid values", function () {
+      it("should throw with null", function () {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        assert.throws(() => keyInfoToPem(null as any));
+      });
+
+      it("should throw with false", function () {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
cjbarth, line 107: Instead of disabling, let's use `@ts-expect-error`.

tsconfig.json
```diff
@@ -4,9 +4,9 @@
 
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
-    "target": "es2018" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
+    "target": "es2019" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */,
```
cjbarth, line 7: What is the reason for this change? I'm not against it, but it does seem like something that we should save for a semver-major change.

src/crypto.ts
```diff
@@ -1,48 +1,93 @@
...
+ *
+ * With couple of notes:
+ *  - 'eol' is normalized to '\n'
+ */
+const PEM_FORMAT_REGEX =
+  /^(-----BEGIN [A-Z\x20]{1,48}-----(\r\n|\r|\n){1}.*(\r\n|\r|\n){1}-----END [A-Z\x20]{1,48}-----(\r\n|\r|\n){0,1})$/s;
+const BASE64_REGEX =
+  /^(?:[A-Za-z0-9\+\/]{4})*(?:[A-Za-z0-9\+\/]{2}==|[A-Za-z0-9\+\/]{3}=|[A-Za-z0-9\+\/]{4})$/m; // eslint-disable-line no-useless-escape
+
+export const PemLabel = {
```
cjbarth, line 28: If we have to export it, then it seems it should be in `types.ts`.

cjbarth:
Thanks for all your hard work and the clarity of your code. I see that this patch is almost at 100%, good job! Keep at it!

RopoMen:
@cjbarth is there something that should be done to get this PR into master?
- DeepScan finding is false-positive
- codecov finding is minor

src/saml.ts
```diff
@@ -638,30 +668,18 @@ class SAML {
...
-        });
-    } else if (Array.isArray(this.options.cert)) {
-      checkedCerts = this.options.cert;
-    } else {
-      checkedCerts = [this.options.cert];
+      // Do not cache
+      return await resolveAndParseKeyInfosToPem(this.options);
+    } else if (this.pemFiles.length > 0) {
+      // Return already cached PEM files.
+      return this.pemFiles;
```
cjbarth, line 677: This is part of the new functionality that you've added that is completely uncovered by tests. Can you please add one here to make sure that caching is working as intended?

src/pem-label.enum.ts
```diff
@@ -0,0 +1,7 @@
+export const PemLabel = {
```
cjbarth, line 1: What was the reasoning behind putting this type in its own file instead of with all the other types?

test/tests.spec.ts
```diff
@@ -701,7 +701,7 @@ describe("node-saml /", function () {
         const base64xml = Buffer.from(xml).toString("base64");
         const container = { SAMLResponse: base64xml };
         const samlObj = new SAML({
-          cert: "-----BEGIN CERTIFICATE-----" + TEST_CERT + "-----END CERTIFICATE-----",
+          cert: `-----BEGIN CERTIFICATE-----\n${TEST_CERT}\n-----END CERTIFICATE-----`,
```
cjbarth, line 704: This technically changes the test. Will the code still work if there isn't a newline after the opening segment or before the closing segment? I know that is non-standard, but, since it is supported that way in a test, we should probably continue supporting it until a semver-major.

RopoMen:
@cjbarth Hi, is there something that you are expecting me to work with this PR? Or do you have time allocation issues with this PR? I would like to work to get this PR finished.

cjbarth:
It seems that my previous PR comments are unresolved. Why not start there? I'm also interested in addressing those regex warnings, but I need more time to review what can be done there.

test/crypto.spec.ts
cjbarth: We don't want to disable checking, we want to inform the linter that an error should be here.

RopoMen:
DeepScan stuff fixed, I'll check if I can get coverage a bit better.

RopoMen:
@cjbarth Hi, I resolved most of the conversations on couple of weeks ago, but there are still those two open conversations. For now I need more feedback from you in order to get things moving.

If there are some issues which needs to be addressed in specific way other than those two open conversations, I wish you point out those explicitly where the issue is and how you want it to be fixed so that I can make those fixes.

cjbarth:
I'm sorry for the delay; I was unavailable for a while. I'm back to giving this project more attention. I've read over the changes you've made and they look good. I'll take time to take a second look at my comments that are still open and see if they are still valid or if I need to add more details.

cjbarth:
@RopoMen , I've pushed up a lot of changes to improve this code. We don't want to export code just for testing, so many tests were changed, or even removed, for that reason. Many tests were redundant.

I see the point about the `codecov` tool. I'll merge if you agree with the changes I've made. I look forward to your feedback.

RopoMen:
> @RopoMen , I've pushed up a lot of changes to improve this code. We don't want to export code just for testing, so many tests were changed, or even removed, for that reason. Many tests were redundant.
> 
> I see the point about the `codecov` tool. I'll merge if you agree with the changes I've made. I look forward to your feedback.

Hi @cjbarth 
You can merge this if you are done. I don't mind about the changes, main point is more important.

Thank you to be active again 😁

PR #235, opened 2022-12-30
feat: add public getAuthorizeMessage method

### LINKED ISSUE #94: [ENHANCE] Public getSAMLRequestValue method

I'm using node-saml as part of my `Remix` web stack. However, I don't want to directly use the `getAuthorizeFormAsync` as it means I can't customise the form at all and just generally makes working with the library more difficult then I feel it needs to be.

A getSAMLRequestValue (or similarly named) method on the main `SAML` class that I can use to generate the Base64 encoded SAMLRequest value, to pass into my own form.

I'm doing the following as a workaround currently which works but isn't ideal:
```ts
 const authForm = await saml.getAuthorizeFormAsync("");
  // Extract the SAMLRequest value with regex
  const SAMLRequestValue = authForm
    .split(`name="SAMLRequest" value="`)[1]
    .split(`"`)[0];
```

I would love to contribute to the library and if I can I'll get a PR raised to propose a solution.

aykutbulca:
# Description

Added a new method that generates and returns the SAML authorization message without including the `Html form` details.

We don't directly use the generated authorization form value (`getAuthorizeFormAsync`) and extract the `SAMLRequest` from the form values to generate our SAML request message. The use case is similar to the one explained in [this issue](https://github.com/node-saml/node-saml/issues/94) (Fixes #94). 

Since it was considered a valuable addition to the library, I extracted the required logic from the `getAuthorizeFormAsync` method into a new one (`getAuthorizeMessageAsync`), added tests for the `SAMLRequest` and additional parameters.

# Checklist

- Issue Addressed: [X]
- Link to SAML spec: [ ]
- Tests included? [X]
- Documentation updated? [ ]

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -515,6 +515,33 @@ class SAML {
     );
   }
 
+  async getAuthorizeMessageAsync(
+    RelayState: string,
+    host?: string
+  ): Promise<querystring.ParsedUrlQueryInput> {
+    assertRequired(this.options.entryPoint, "entryPoint is required");
+
+    const request = await this.generateAuthorizeRequestAsync(this.options.passive, true, host);
+    let buffer: Buffer;
+    if (this.options.skipRequestCompression) {
+      buffer = Buffer.from(request, "utf8");
+    } else {
+      buffer = await deflateRawAsync(request);
+    }
+
+    const operation = "authorize";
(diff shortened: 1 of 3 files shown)


cjbarth (approved):
Very nice test suite!

PR #262, opened 2023-04-03
Allow to configure optional logger function to log SAML requests/resp…

RajuTanchak:
# Allow to configure optional logger function to log SAML operations

We have migrated our application's Login component away from Java/Spring to Typescript with `passport-saml`.

One facet of the Spring SAML libraries which we found very useful when debugging SAML config issues in production was the ability to configure the library to ouput logs for key SAML operations. [This was introduced in](https://github.com/spring-attic/spring-security-saml/issues/63) with a comment in the Spring code that said "[Implementations are supposed to log significant SAML operations](https://github.com/spring-attic/spring-security-saml/blob/main/core/src/main/java/org/springframework/security/saml/log/SAMLLogger.java#L22)."

We have been unable to find a reference to this "supposed to" in the formal SAML specifications, but as mentioned it has been a very useful capability in the past, so we thought to propose this approach as a way to address the issue described in https://github.com/node-saml/passport-saml/issues/279.

# Checklist:

- Issue Addressed: [node-saml/passport-saml/issues/279]
- Link to SAML spec: [ ]
- Tests included? [Yes]
- Documentation updated? [Yes]

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ const saml = new SAML(options);
 - `generateUniqueId`: optional function which will be called to generate unique IDs for SAML requests.
 - `scoping`: An optional configuration which implements the functionality [explained in the SAML spec paragraph "3.4.1.2 Element <Scoping>"](https://docs.oasis-open.org/security/saml/v2.0/saml-core-2.0-os.pdf). The config object is structured as following:
 - `signMetadata`: if true, adds a signature to the generated Service Provider metadata. `privateKey` must be set to use this option.
+- `log`: optional logger function to log SAML request and response.
 
 ```javascript
 {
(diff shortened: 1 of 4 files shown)


cjbarth:
I'm not against this, but I worry that it could be left on in production and then create some issues with log volume and security. Should it also be behind a debug flag? Can be leverage the `debug()` function that we already have, or perhaps migrate our one usage of `debug()` to a system unified with what you have here?

cjbarth:
@RajuTanchak , what do you think of @srd90 's comments?

cjbarth:
I'm going close this as there has been no response from the author, there are significant arguments against this, and there are merge conflicts.

PR #266, opened 2023-04-10
bump xml2js to 0.5.0 dependency for security vulnerability fix

m-salman-afzal:
# Description
node-saml depends upon xml2js package which was recently diagnosed with a security vulnerability as stated in [here](https://github.com/advisories/GHSA-776f-qx25-q3cc). I've just updated the package.json and lock file to reflect changes.

# Checklist:

- Issue Addressed: [x]
- Link to SAML spec: [ ]
- Tests included? [ ]
- Documentation updated? [ ]

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1104,8 +1104,8 @@
       }
     },
     "@types/xml2js": {
-      "version": "0.4.11",
-      "resolved": "https://registry.npmjs.org/@types/xml2js/-/xml2js-0.4.11.tgz",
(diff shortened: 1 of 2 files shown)


cjbarth:
Since the dependency is updated, this PR isn't strictly needed. However, I'd love to see the signature validation issues that @srd90 pointed out addressed. I'll keep my eye out for such a PR (could even be this one). @m-salman-afzal , would you be in a position to do that?

PR #278, opened 2023-05-26
Fix spelling and normalize naming

cjbarth:
We had several spelling mistakes, even in variable names, throughout the code-base. This PR corrects that, which means there are some renamed variables, so this is a major change.


cjbarth:
@markstos , I note that our test coverage went down. I'll add some more tests to cover that late. No tests were removed, but as code got cleaned up, I think the ratio became more unfavorable.

markstos:
I reviewed this again and found that while some variables were renamed internally, I don't see any *user-facing* variables were renamed.

So if people are landing here looking for breaking changes to address, I don't see any user-facing breaking changes here.

cjbarth:
If you use TypeScript, it is a breaking change, e.g. `RacComparison`.

PR #285, opened 2023-06-14
Add tests for XML parsing with comments

cjbarth:
These tests don't improve coverage, but they do make sure that there is no regression in dependent libraries, specifically around XML parsing.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,6 +2,7 @@
   "cSpell.words": [
     "ADFS",
     "apos",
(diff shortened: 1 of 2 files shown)


cjbarth:
@markstos , @srd90, I wrote this code a while ago, and I found it while cleaning up. It was still compatible with our code-base, so I've put up a PR. However, I don't want to include code that doesn't add value as that just adds maintenance overhead. What are your thoughts about including this?

PR #289, opened 2023-06-28
Defend against undefined NotOnOrAfter

joshgummersall:
The [docs](https://github.com/node-saml/node-saml#saml-response-validation---notbefore-and-notonorafter) explain that NotOnOrAfter and NotBefore will only be validated if they exist in the SAML response. I found that this produces a runtime exception that `NotOnOrAfter` is undefined, thrown [here](https://github.com/node-saml/node-saml/blob/master/src/saml.ts#L1340). This was the simplest fix I could find that does not affect any existing tests. Thanks for your time & a great library!

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -1022,7 +1022,10 @@ class SAML {
           const _confirmData = _subjectConfirmation.SubjectConfirmationData?.[0];
           if (_confirmData?.$) {
             const subjectNotBefore = _confirmData.$.NotBefore;
+
             const subjectNotOnOrAfter = _confirmData.$.NotOnOrAfter;
+            if (subjectNotOnOrAfter == null) return true;
+
             const maxTimeLimitMs = this.calcMaxAgeAssertionTime(
               this.options.maxAssertionAgeMs,
               subjectNotOnOrAfter,
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -1024,13 +1024,14 @@ class SAML {
             const subjectNotBefore = _confirmData.$.NotBefore;
 
             const subjectNotOnOrAfter = _confirmData.$.NotOnOrAfter;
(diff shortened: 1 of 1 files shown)


src/saml.ts
```diff
@@ -1022,12 +1022,16 @@ class SAML {
...
             const subjectNotBefore = _confirmData.$.NotBefore;
+
             const subjectNotOnOrAfter = _confirmData.$.NotOnOrAfter;
-            const maxTimeLimitMs = this.calcMaxAgeAssertionTime(
-              this.options.maxAssertionAgeMs,
-              subjectNotOnOrAfter,
-              assertion.$.IssueInstant
-            );
+
+            const maxTimeLimitMs = subjectNotOnOrAfter
```
cjbarth, line 1028: Correction to this logic should be in `this.calcMaxAgeAssertionTime()`, not here. We should be able to handle bad data coming in and still return a proper value per the existing logic.

cjbarth (changes requested):
We also need a test to make sure that this doesn't break again.

cjbarth:
@joshgummersall, thanks for your PR. Please note the comments and we'll do our best to help you get this code landed.

src/saml.ts
joshgummersall: @cjbarth, `checkTimestampsValidityError` properly handles `maxTimeLimitMs` that is `undefined`, would you say it's acceptable to return `undefined` from `checkTimestampsValidityError`?

joshgummersall:
@srd90 yeah, my reading of the spec you shared is the same. It seems that `NotBefore` is optional, but `NotOnOrAfter` is mandatory. I'm working with an internal IDP that I think I can submit a fix for instead or trying to land this. Thanks for sharing that resource!

joshgummersall:
All squared away, thanks again for the pointers!

PR #295, opened 2023-07-07
Move @types/* to devDependencies

Kakulukian:
# Description

This PR updates the package.json file for the project to move the @types/ packages from dependencies to devDependencies. I have custom typing for @types/passport, and this package will overwrite it

PR #296, opened 2023-07-13
Remove dependency on Passport types

cjbarth:
As mentioned in #295, there shouldn't be any need to depend on Passport types. This removes that requirement.


cjbarth:
This should address #295.

PR #314, opened 2023-08-09
Enforce valid setting for validateInResponseTo

adamandreasson:
# Description

This PR addresses the potential security vulnerability I raised in the passport-saml repo here: https://github.com/node-saml/passport-saml/issues/874

* Enforce only valid settings are passed to validateInResponseTo and throw an error if any other value is passed
* Added tests for the new enforcement

A separate PR will be submitted in the passport-saml repo that updates the documentation to reflect these changes.

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -150,6 +150,10 @@ class SAML {
       racComparison: ctorOptions.racComparison ?? "exact",
     };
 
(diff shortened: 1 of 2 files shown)


cjbarth (approved):

cjbarth:
I'd like to see another PR that sets the default to the more secure `always`.

PR #316, opened 2023-08-16
Support multiple callback urls

### LINKED ISSUE #315: [ENHANCE] Support multiple AssertionConsumerService properties in metadata
It would be great to be able to be able to add multiple `AssertionConsumerService` properties when generating the metadata.

The `generateServiceProviderMetadata` should support passing multiple `callbackUrl`.
For backwards compatibility this could just mean supporting passing an `string[]` to `callbackUrl`.

dertieran:
# Description
Closes #315 
 
This PR adds support for passing multiple callback urls to generate multiple [`AssertionConsumerService`](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf) properties. 
 
Right now this is done by supporting passing an `string[]` instead of just a `string` for the `callbackUrl`.

Not sure if this is the cleanest solution or if this should be done through another option, something more like `additionalCallbackUrls`.

# Checklist:

- [x] Issue Addressed
- [x] Link to SAML spec
- [x] Tests included?
- [ ] Documentation updated?

### DIFF
--- a/src/metadata.ts
+++ b/src/metadata.ts
@@ -2,7 +2,6 @@ import { stripPemHeaderAndFooter } from "./crypto";
 import {
   isValidSamlSigningOptions,
   ServiceMetadataXML,
-  XMLObject,
   GenerateServiceProviderMetadataParams,
 } from "./types";
 import { assertRequired, signXmlMetadata } from "./utility";
@@ -129,12 +128,15 @@ export const generateServiceProviderMetadata = (
     metadata.EntityDescriptor.SPSSODescriptor["@WantAssertionsSigned"] = true;
   }
 
-  metadata.EntityDescriptor.SPSSODescriptor.AssertionConsumerService = {
-    "@index": "1",
-    "@isDefault": "true",
-    "@Binding": "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST",
-    "@Location": callbackUrl,
-  } as XMLObject;
+  const callbackUrls = Array.isArray(callbackUrl) ? callbackUrl : [callbackUrl];
+  metadata.EntityDescriptor.SPSSODescriptor.AssertionConsumerService = callbackUrls.map(
+    (url, index) => ({
+      "@index": String(index + 1),
+      "@isDefault": index === 0 ? "true" : "false",
+      "@Binding": "urn:oasis:names:tc:SAML:2.0:bindings:HTTP-POST",
+      "@Location": url,
+    }),
+  );
 
   let metadataXml = buildXmlBuilderObject(metadata, true);
   if (params.signMetadata === true && isValidSamlSigningOptions(params)) {
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -224,7 +224,10 @@ class SAML {
     }
 
     if (!this.options.disableRequestAcsUrl) {
-      request["samlp:AuthnRequest"]["@AssertionConsumerServiceURL"] = this.options.callbackUrl;
+      const callbackUrl = Array.isArray(this.options.callbackUrl)
+        ? this.options.callbackUrl[0]
+        : this.options.callbackUrl;
+      request["samlp:AuthnRequest"]["@AssertionConsumerServiceURL"] = callbackUrl;
     }
 
     const samlAuthnRequestExtensions = this.options.samlAuthnRequestExtensions;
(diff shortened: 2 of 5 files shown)


markstos:
Could you update the documentation as well?

cjbarth:
> introduce different SP metadata for different domains so that you could be sure that same domain name is used for SLO as is used for ACS

This is the correct solution. You simply pass difference configuration data to the routines that generate the metadata based on the domain the request for metadata is coming from and everything works like it should. This is the method I've used in the past.

cjbarth:
> at least one SP implementation - passport-saml - doesn't work properly

@srd90 , I can't express enough how much I appreciate your involvement with this project. I've spent a little time reviewing our previous discussions on this matter, specifically your [comment](https://github.com/node-saml/passport-saml/issues/419#issuecomment-1567565893) and I see that, while `node-saml` and `passport-saml` don't come out-of-the-box with support for SLO, they do support it. It is only broken if it is implemented incorrectly. In fact, due to the flexibility of these libraries, there is no way to support it natively OOTB. Do I understand all that correct?

If I'm misunderstanding things, please help me to see where my comprehension fails me. Otherwise, I think it is important that we make sure we're stating clearly, not that SLO is broken, but that it isn't supported OOTB, just like the [README](https://github.com/node-saml/passport-saml/pull/862) says. If the README is wrong in this sensitive area, we certainly want to correct it.

dertieran:
Not sure if it is better to discuss this here or in the issue.

@srd90 is right my use case would be that I have on endpoint to get the metadata of my SP which supports a domain wide login to different services, each on their own subdomain.
So e.g. `https://example.com/saml/metadata` to get the metadata and the user is able to log in from `https://a.example.com`, `https://b.example.com` and so on.

And most of the users configure their IdP manually anyway, so this normally isn't an issue.
But I had one user just took the metadata and added it to their IdP and tried to log in from `https://b.example.com`.

I already had implemented, what you also suggested, that calling the `metadata` endpoint from a subdomain would adjust the `callbackUrl` to match that domain.

But I thought it would be nice if the user could just take the metadata file and upload it to their IdP.

<details>
<summary>So e.g. in Azure AD you could just take the metadata.xml from the test and upload it.</summary>

image
</details>

I can see that this isn't the main usecase for `node-saml` and I wasn't really aware of the implications.
Maybe it would even be better to have this in a separate option *(like `additionalCallbackUrls`)* instead of overlaoding the `callbackUrl`.

But I can also see that this might be a niche usecase and if it complicates other things I would be also fine with dropping this PR.

cjbarth:
> But I thought it would be nice if the user could just take the metadata file and upload it to their IdP.

They can, and that is what I've done. Just have the request for metadata generate the metadata different depending on the domain they hit it from. Or, just make the metadata request path something like `https://example.com/saml/metadata/a` and `https://example.com/saml/metadata/b`, etc.

cjbarth:
Is this still an open issue based on the feedback provided @dertieran

dertieran:
@cjbarth sorry for the delayed response.
Using different endpoints/domains for the metadata is what I already implemented.
But the user then has multiple metadata files and would need to create multiple *"applications"* in e.g. Azure AD.
With the single metadata file they could just take it and upload it, as shown in the screenshot.

But if this isn't completely spec compliant or produces other issues I'm also fine to close this PR/Issue.
Because theoretically this is just relevant for the metadata for me, and I could just generate them without this library.

cjbarth:
@dertieran  Are you saying that all those domains would point to the same session store, the same application? i.e. if a SLO request were to hit any of the domains, it would behave just as if it hit any of the other domains?

dertieran:
Yes, at least for me, these are the same servers, but I don't think they would have to be.

Essentially I have a service where the user has one account *(and one SSO configuration)*, but can use different services *(depending on the subdomain)*.
So a user might need to konfigure all subdomains/services in one *"application"* in the IdP so it will have the same configuration.

But again, feel free to close this if this seems to be a niche use case.
And the important part *(logging in the user)* works as expected, I just wanted to give the user a better `metadata.xml` 😃

cjbarth:
This is probably why subdomains for shared services are no longer the industry standard for sites sharing a common security context. I might suggest that you have one login domain and allow that token to be used for the other domains if you can't combine the sites. In reality, the biggest reason to use different subdomains is to separate security contexts. What your after is the opposite of that, so I'd say it is niche, so I'll close this.

e.g. browse to https://maps.google.com/ and note that you are redirected to https://www.google.com/maps

PR #318, opened 2023-08-18
Fixed test failure

jindazhao01:
# Description

Deleted duplicated test in ```validateRedirect()```

```  210 passing (561ms)```


jindazhao01:
@cjbarth @markstos not sure why I was denied permissions to push to the repo itself, but tagging you here as this is the fix

Issue #70, opened 2022-04-27
[ENHANCE] Add test coverage for every exception we throw


cjbarth:
In order to continue ensuring that invalid conditions, which would represent security issues, are correctly detected, we need some better test coverage. Currently, we have many places where we throw exceptions if there are problems with SAML processing. However, we don't have test coverage for these conditions.

It is currently possible to modify code to allow an unsafe condition through as valid SAML and still have all the existing tests pass. It is very difficult to catch such cases in code review. Thus, we need to add tests that will cover every exception that we throw so that we can ensure these invalid cases are always detected correctly.

PR #327, opened 2023-09-21
Add test coverage for initialize() of saml.ts

kg0r0:
# Description

I worked on some of the issue https://github.com/node-saml/node-saml/issues/70 .
Perhaps this issue could be addressed in stages.

# Checklist:

- Issue Addressed: [x]
- Link to SAML spec: N/A
- Tests included? [x]
- Documentation updated? N/A

### DIFF
--- a/test/samlTests.spec.ts
+++ b/test/samlTests.spec.ts
@@ -448,5 +448,19 @@ describe("saml.ts", function () {
         );
       });
     });
+    describe("initialize", function () {
+      it("should throw a error when SamlOptions is not set", function () {
+        expect(() => {
+          const samlObj = new SAML({
+            callbackUrl: "http://localhost/saml/consume",
+            cert: FAKE_CERT,
+            issuer: "onesaml_login",
+            audience: false,
+          });
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          samlObj.initialize(undefined as any);
+        }).to.throw("SamlOptions required on construction");
+      });
+    });
   });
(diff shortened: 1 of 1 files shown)


cjbarth:
Thank you very much for contributing this. Can you please address the problems with linting?

kg0r0:
> Thank you very much for contributing this. Can you please address the problems with linting?

Hi @cjbarth,
Sorry I did not notice the eslint error.

It seems that this repository allows the use of  ``// eslint-disable-next-line @typescript-eslint/no-explicit-any ``. Is this correct?

https://github.com/node-saml/node-saml/blob/cbe102ab99346f5282dfb2ab629d1ffac6504c6f/src/types.ts#L45-L46

I think sometimes it is necessary to pass a type check in order to test every exceptions, and a type assertion using any is used for this purpose.
If the above is ok, I can rebase and push the changes again to wrap up the commits.

cjbarth:
For test especially, that is true. Did you try `unknown`?

kg0r0:
> For test especially, that is true. Did you try `unknown`?

Okay, this is test code so I will push that change.
I also tried `unknown` but this did not seem to work.

```
Argument of type 'unknown' is not assignable to parameter of type 'SamlConfig'.
  Type 'unknown' is not assignable to type 'Partial<SamlOptions>'.ts(2345)
```

cjbarth:
Ok, then put in the comment :)

cjbarth (approved):

Issue #336, opened 2023-12-05
Generate SP metadata without knowing IdP cert


nwalters512:
**Is your feature request related to a problem? Please describe.**

I use the `MultiSamlStrategy` Passport strategy. I'd like to provide metadata URLs to our IdPs, which I can build with `MultiSamlStrategy#generateServiceProviderMetadata`.

This works fine if I know the IdP's cert and can have `getSamlOptions` return an option object with `cert`. However, during initial setup, I may not immediately know the IdP's certificate. I should still be able to generate metadata for them, as (as far as I know) that metadata doesn't depend on their own certificate. However, if `getSamlOptions` returns `cert: null` (because I don't yet know it), the following assertion fails:

https://github.com/node-saml/node-saml/blob/e691ccf39b9f3521268fc31b8b8f099beae4654c/src/saml.ts#L98

**Describe the solution you'd like**

Since the XML metadata doesn't depend on the IdP's cert, I'd like to be able to generate it before I do in fact know their cert.

**Describe alternatives you've considered**

When someone asks for our metadata URL, I can of course say "no, _you_ give me your metadata URL first so I can get your cert". However, it would be very nice to be able to always provide it first.

cjbarth:
Just at first glance, i believe this is supposed to be your public cert. Otherwise, I agree with you that you should be able to generate metadata without anything from the IdP.

nwalters512:
@cjbarth The `node-saml` docs state that this should be the IdP's public cert:

> `cert`: the IDP's public signing certificate used to validate the signatures of the incoming SAML Responses

@srd90 thank you for the link to that discussion, I'm very much in agreement that having a certificate configured is critical for security and I'm not trying to question that.

It didn't occur to me that I could provide _any_ string for the cert, but yes, given that the constructor currently only validates that it's a string and not that it's a well-formed and valid certification, that should be safe. I'd be a little worried about future changes that move such validation to the constructor. If such a change is ever considered, this is a use case that should be considered with it.

Using `generateServiceProviderMetadata` is also a good idea, since it seems like that takes an object that basically matches what `getSamlOptions` will return. However, that doesn't appear to be a public API (it's not exported from `src/index.ts`. @cjbarth, do you think you or another maintainer would accept a PR that officially makes `generateServiceProviderMetadata` part of the API?

cjbarth:
I have no objection to a PR that makes `metadata/generateServiceProviderMetadata()` part of the API as long as there is a test for it.

PR #337, opened 2023-12-06
Export generateServiceProviderMetadata

nwalters512:
# Description

This PR makes the change proposed in #336 and directly exports `generateServiceProviderMetadata`. This is useful for generating metadata without constructing a complete and valid `SAML` instance.

For this function to be maximally useful on its own, I made some changes so that it would behave the same as if it were used on an instance of `SAML`. That is, for some set of options, `new SAML(options).generateServiceProviderMetadata() === generateServiceProviderMetadata(options)` should be true. This mostly entailed making additional options optional and providing defaults that matched those set in `SAML#initialize()`.

# Checklist:

- Issue Addressed: #336
- Link to SAML spec: N/A
- Tests included? [x]
- Documentation updated? [x]

### DIFF
--- a/src/metadata.ts
+++ b/src/metadata.ts
@@ -16,12 +16,13 @@ export const generateServiceProviderMetadata = (
     issuer,
     callbackUrl,
     logoutCallbackUrl,
-    identifierFormat,
-    wantAssertionsSigned = true,
     decryptionPvk,
     privateKey,
     metadataContactPerson,
     metadataOrganization,
+    // The below defaults are selected to match the defaults in the `SAML` class.
+    identifierFormat = "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress",
+    wantAssertionsSigned = true,
     generateUniqueId = generateUniqueIdDefault,
   } = params;
 
--- a/src/metadata.ts
+++ b/src/metadata.ts
@@ -8,6 +8,7 @@ import {
 import { assertRequired, signXmlMetadata } from "./utility";
 import { buildXmlBuilderObject } from "./xml";
 import { generateUniqueId as generateUniqueIdDefault } from "./crypto";
+import { DEFAULT_IDENTIFIER_FORMAT, DEFAULT_WANT_ASSERTIONS_SIGNED } from "./constants";
 
 export const generateServiceProviderMetadata = (
   params: GenerateServiceProviderMetadataParams,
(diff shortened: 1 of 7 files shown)


src/types.ts
```diff
@@ -84,6 +84,7 @@ export type CertCallback = (
 export interface MandatorySamlOptions {
   cert: string | string[] | CertCallback;
   issuer: string;
+  callbackUrl: string;
```
nwalters512, line 87: I noticed while writing the test that `callackUrl` was incorrectly marked as optional given that there's a runtime check to make sure it's provided:

https://github.com/node-saml/node-saml/blob/e691ccf39b9f3521268fc31b8b8f099beae4654c/src/saml.ts#L96

I acknowledge that this isn't under the original scope of the PR, I'd be happy to move it to another one if requested.

src/metadata.ts
```diff
@@ -15,13 +16,14 @@ export const generateServiceProviderMetadata = (
...
     logoutCallbackUrl,
-    identifierFormat,
-    wantAssertionsSigned,
     decryptionPvk,
     privateKey,
     metadataContactPerson,
     metadataOrganization,
-    generateUniqueId,
+    // The below defaults are selected to match the defaults in the `SAML` class.
+    identifierFormat = "urn:oasis:names:tc:SAML:1.1:nameid-format:emailAddress",
```
cjbarth, line 24: I'd prefer to keep all the defaults in a single location. If they need to be duplicated here, then perhaps we should put a run-time check to make sure they are populated and leave the onus on the one calling this method directly to know what their doing.

nwalters512: Is there any particular way you'd like me to share these defaults? Maybe declare them in a constant somewhere? Is there precedent for this?

I'd rather not leave this to the caller; if I'm not explicitly configuring `identifierFormat` for instances of `SAML`, I wouldn't expect to have to configure them here either.

cjbarth: I think I might be OK with a set of constants in a new file: `constants.ts`. I see we've already polluted the `types.ts` file a little with a class that should probably be in `utilities.ts`, but I don't want that to be a pattern. It also doesn't seem to be a great fit for `utilities.ts`. I'm interested in your thoughts.

I really want to make sure that the next guy (or future me) doesn't think there is just one place for defaults and forgets to change something.

nwalters512: I think your motivation is spot on. I think a `constants.ts` file with exports like `DEFAULT_IDENTIFIER_FORMAT` and `DEFAULT_WANT_ASSERTIONS_SIGNED` would be a good idea; I'll make that change if you're good with it!

nwalters512: @cjbarth done in commit d38a77d.

nwalters512:
Looks like some tests were failing in CI. However, they also fail on `master` both locally and on CI, so I'm guessing that isn't anything to do with my changes.

cjbarth:
@nwalters512 , I think everything checks out now. Can you give this a final once-over to make sure that the merge I did still looks good? I want to do a release and this code is close to landing, so I want to include it.

nwalters512:
@cjbarth thanks for getting this up to date with master. This looks ready to go to me!

Issue #339, opened 2023-12-21
[BUG] From multiple Audience variants, only the first one gets looked at


catamphetamine:
We're using `node-saml` as a "plugin" for SAML authentication at our company.
We've recently had an issue with one of our clients who is sending multiple `Audience`s in one `AudienceRestriction`:

```xml
<saml2:Conditions NotBefore="2023-12-20T20:44:20.578Z" NotOnOrAfter="2023-12-20T20:49:20.577Z">
    <saml2:AudienceRestriction>
        <saml2:Audience>
            acadeum
        </saml2:Audience>
        <saml2:Audience>
            https://students.acadeum.com
        </saml2:Audience>
    </saml2:AudienceRestriction>
</saml2:Conditions>
```

At first, we thought that such input is invalid, but having googled for it, looks like it is valid according to SAML specification:
https://stackoverflow.com/questions/43082519/does-ping-support-multiple-audience-restriction-values-in-saml

Furthermore, after looking at `node-saml` source code, we can see how it simply discards any "audience"s except the first one:
https://github.com/node-saml/node-saml/blob/e691ccf39b9f3521268fc31b8b8f099beae4654c/src/saml.ts#L1229-L1237

```js
if (restriction.Audience[0]._ !== expectedAudience) {
  return new Error(
    "SAML assertion audience mismatch. Expected: " +
      expectedAudience +
      " Received: " +
      restriction.Audience[0]._,
  );
}
return null;
```

It's not clear why it was written the way it is, but we think that the code above should be corrected:

```js
let i = 0;
while (true) {
  if (restriction.Audience[i]._ === expectedAudience) {
    break;
  }
  if (!restriction.Audience[i + 1]) {
    return new Error(
      "SAML assertion audience mismatch. Expected: " +
        expectedAudience +
        " Received: " +
        restriction.Audience[i]._,
    );
  }
  i++;
}
return null;
```

We've submitted the code patch above in the form of a [pull request](https://github.com/node-saml/node-saml/pull/340).

PR #340, opened 2023-12-21
Fixes `node-saml` not checking all `Audience`s in an `AudienceRestriction`

catamphetamine:
### Description

See the issue description: https://github.com/node-saml/node-saml/issues/339

### SAML Specification

https://docs.oasis-open.org/security/saml/v2.0/sstc-saml-approved-errata-2.0-cd-02.pdf

> E46: AudienceRestriction Clarifications

> Note that multiple `<AudienceRestriction>` elements MAY be included in a single assertion, and each MUST be evaluated independently. The effect of this requirement and the preceding definition is that **within a given `<AudienceRestrictions>`, the `<Audience>` elements form a disjunction (an "OR")** while multiple `<AudienceRestrictions>` elements form a conjunction (an "AND").

### Test

Included.

### Documentation Changes

Not required.

# Checklist:

- Issue Addressed: [x]
- Link to SAML spec: [x]
- Tests included? [x]
- Documentation updated? [ ]


src/saml.ts
```diff
@@ -1226,13 +1226,23 @@ class SAML {
...
-        if (restriction.Audience[0]._ !== expectedAudience) {
-          return new Error(
-            "SAML assertion audience mismatch. Expected: " +
-              expectedAudience +
-              " Received: " +
-              restriction.Audience[0]._,
-          );
+        const receivedAudiences = [];
+        let i = 0;
+        while (i < restriction.Audience.length) {
```
cjbarth, line 1231: Why can't you use `.some()`?

catamphetamine: @cjbarth I didn't know it could do that. Rewrote using `.some()`.

src/saml.ts
```diff
@@ -1226,12 +1226,12 @@ class SAML {
...
           return new Error("SAML assertion AudienceRestriction has no Audience value");
         }
-        if (restriction.Audience[0]._ !== expectedAudience) {
+        if (!restriction.Audience.some(audience => audience._ === expectedAudience)) {
           return new Error(
             "SAML assertion audience mismatch. Expected: " +
               expectedAudience +
               " Received: " +
-              restriction.Audience[0]._,
+              restriction.Audience.map(_ => _._).join(", "),
```
cjbarth, line 1234: This seems crazy, but could we change this so that we don't use `_` as an argument to a function that is used. Typically `_` as an argument would indicate that it is throw-away or unused.

```suggestion
              restriction.Audience.map(a => a._).join(", "),
```

catamphetamine:
@cjbarth Thx, updated the variable name as per your suggestion.

src/saml.ts
```diff
@@ -1226,12 +1226,12 @@ class SAML {
         if (!restriction.Audience || !restriction.Audience[0] || !restriction.Audience[0]._) {
           return new Error("SAML assertion AudienceRestriction has no Audience value");
         }
-        if (restriction.Audience[0]._ !== expectedAudience) {
+        if (!restriction.Audience.some(audience => audience._ === expectedAudience)) {
```
cjbarth, line 1229: Would it be more logical to say "if every item doesn't match then throw an error" vs. what you have here which is "if you didn't find some item that matches, throw an error".

```suggestion
        if (restriction.Audience.every(audience => audience._ !== expectedAudience)) {
```

cjbarth:
Please address the build failures. You can check locally by running `npm test`.

catamphetamine:
@cjbarth 

> Please address the build failures. You can check locally by running npm test.

There're 4 tests that don't pass and I haven't touched any of those so dunno.

```
  Comparison {
+   message: 'error:1E08010C:DECODER routines::unsupported'
-   message: 'Failed to read private key'
  }
```

cjbarth:
Does a checkout of `master` pass?

catamphetamine:
@cjbarth I also could speculate that those tests might somehow be broken in the main branch.

I won't be able to tell you if the tests pass locally on the master branch because on my Windows machine a lot of tests don't pass, most likely due to "CR LF" or something.

Perhaps you could checkout this branch and see if the tests pass on your machine.

image

cjbarth:
There is something else going on than a problem on `master`. I don't have time to look at it now, but, if you're on Windows I suggest turning off the automatic line-ending changing feature in git. I also suggest working in WSL. I'm also on Windows.

image

cjbarth:
You're branch is passing tests, so it must be something with a dependency (we check updating dependencies during CI so we can catch stuff earlier). I'll land this and then see what's up.

catamphetamine:
@cjbarth Cool, thanks.

PR #341, opened 2024-01-09
Upgrade to latest version of xml-crypto

cjbarth:
The latest version of xml-crypto has many breaking changes and a few new patterns that we can leverage. This PR makes use of this newly updated library.

Note: a parameter that was optional is no longer so: `signatureAlgorithm`

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ const saml = new SAML(options);
 - `cert`: the IDP's public signing certificate used to validate the signatures of the incoming SAML Responses, see [Security and signatures](#security-and-signatures)
 - `privateKey`: see [Security and signatures](#security-and-signatures).
 - `decryptionPvk`: optional private key that will be used to attempt to decrypt any encrypted assertions that are received
-- `signatureAlgorithm`: optionally set the signature algorithm for signing requests, valid values are 'sha1' (default), 'sha256', or 'sha512'
+- `signatureAlgorithm`: valid values are 'sha1', 'sha256', or 'sha512'
 - `digestAlgorithm`: optionally set the digest algorithm used to provide a digest for the signed data object, valid values are 'sha1' (default), 'sha256', or 'sha512'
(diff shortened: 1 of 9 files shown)


cjbarth:
@LoneRifle , I'd like to see what you have to say about this since you've been involved in `xml-crypto` and `is-dom-node`.

cjbarth:
@srd90 , I've replicated your findings and appreciate your comment. I guess, in harmony with the comments in #399, your concern here is that, while we "fixed" this issue in xml-crypto@5 by making sure there is no KeyInfo checking by default, we undid that here by setting a default. Does that sound correct?

cjbarth:
@srd90 After more closely reviewing the code and your comments, it does seem like, per the SAML spec, which uses out-of-band key exchange as part of the [metadata](https://docs.oasis-open.org/security/saml/Post2.0/sstc-saml-tech-overview-2.0.html#:~:text=metadata%20defines%20a%20way%20to%20express%20and%20share%20configuration%20information%20between%20saml%20parties.), we shouldn't be leveraging the `<KeyInfo />` element. So, while appropriate for `xml-crypto`, it should be turned off for `node-saml`.

I'd like to see what @LoneRifle and @markstos  have to say about this before I land this.

cjbarth:
@srd90 , @LoneRifle , I've incorporated all your suggestions and am ready to land this. @markstos , did you want to give this a second set of eyes?

markstos (approved):
This looks good to me.  Thanks for working on this!

I presume there will be a major version bump due to making an optional argument required. Some folks may be prefer we default to a newer algorithm instead, but I think there's also some value in requiring this to be explicit.

The commit history could a bit cleaner by merging the `wip`commit with another and merging a couple of the node compat commits as well.

cjbarth:
@markstos , thanks for your review. I'll do a squash merge anyway, so the commits don't matter. And, yes, this is in preparation for a (long overdue) major version bump.

markstos:
I found more of @cjbarth rationale in the xml-crypto bug tracker:

> ... Where we can't choose a default that is secure and works for everyone, we choose no default and make things throw if the consumer doesn't make a decision.

Makes sense to me and is aligned with the general approach here of trying to be "secure by default".

PR #343, opened 2024-02-07
Rename `cert` to `idpCert` and `signingCert` to `publicCert`

cjbarth:
To increase clarity and align with `xml-crypto`, `cert` has been renamed to `idpCert` and `signingCert` has been renamed to `publicCert`. The thinking is that the name should clearly indicate that private key material should not be present in this variable and where the cert should come from.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ const saml = new SAML(options);
 - `entryPoint`: identity provider entrypoint (is required to be spec-compliant when the request is signed)
 - `issuer`: issuer string to supply to identity provider
 - `audience`: expected saml response Audience, defaults to value of Issuer (if `false`, Audience won't be verified)
-- `cert`: the IDP's public signing certificate used to validate the signatures of the incoming SAML Responses, see [Security and signatures](#security-and-signatures)
+- `idpCert`: the IDP's public signing certificate used to validate the signatures of the incoming SAML Responses, see [Security and signatures](#security-and-signatures)
 - `privateKey`: see [Security and signatures](#security-and-signatures).
 - `decryptionPvk`: optional private key that will be used to attempt to decrypt any encrypted assertions that are received
 - `signatureAlgorithm`: valid values are 'sha1', 'sha256', or 'sha512'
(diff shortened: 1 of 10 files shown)


markstos:
I love the added clarity. Thank you!

I'm thinking about making the transition gentler for current users.  Here's a behavior suggestion similar to what Mongoose does for some breaking changes:

- In this major version bump, continue to accept `cert`, and when we do so:

- Immediately set the value to `publicCert`, which would be used throughout the rest of the code, as it is here.
- Issue an info-level (or warning?) level warning "`cert` has been renamed to `publicCert` and will be removed in node-saml 6".

Then in the next semver, the code is completely removed. 

Along those lines,  I would consider adding a parenthetical comment to the docs for this version for `publicCert` noting that _"(formerly named `cert`)"_ And also drop that in version 6. 

In the perfect world, everyone reads the Changelog for breaking changes and takes care of everything during an upgrade, but I imagine some folks looking to do the mininum to get a security upgrade done would appreciate some more hints about the change. 

Because I consider these changes "nice to have" rather than "mandatory", I'm still marking the PR as approved.

markstos (dismissed):

salortiz:
My understanding was that `cert` belongs to the IdP (so it can be taken from the `IDPSSODescriptor`, if IdP's metadata is available) while `signingCert` belongs to the SP whose `SPSSODescriptor` metadata will be constructed.

Am I missing something?

cjbarth:
@salortiz thank you for your comment. No you're not missing something, I missed something and got a little too ambitious with my refactoring tools. This does point out why naming things well is so important. I've gone back and fixed this, thought I still can't figure out why the tests fail. If you happen to know why, I'd love a hit... this is a weird one to me.

cjbarth:
> In the perfect world, everyone reads the Changelog for breaking changes and takes care of everything during an upgrade, but I imagine some folks looking to do the mininum to get a security upgrade done would appreciate some more hints about the change.

@markstos , I don't like the idea of having to constantly go back and deal with stuff like this. I'm probably the most active maintainer on this project, and I don't have that much time, and less so as time goes by (I don't use this on a daily basis in my job now, so this is just for support). What do you think of just putting in a deprecation warning in the current major release so if they do a minor release they'll get the warning that something is going to change?

I hesitate to do even that because if you're not reading the changelog, you're probably not reading `console.warn` either. However, if we throw because we're missing the variables that are needed, it should be pretty obvious what little needs to be changed.

salortiz:
@cjbarth, your tests are now failing because commit c5a2cc90 added code to `test/tests.spec.ts` that use the original `cert`.

 See current [tests.spect.ts#L3292](https://github.com/node-saml/node-saml/blob/master/test/tests.spec.ts#L3292) and the [error details](https://github.com/node-saml/node-saml/actions/runs/7831899877/job/21369314125?pr=343#step:5:18)
 
 You need to rebase your branch and change the new test.

PR #367, opened 2024-04-11
Remove unused function arguments

### LINKED ISSUE #381: [ENHANCE] `host` parameter for `getAuthorize*()` is not used at all
`host` parameter is just confusing and since it isn't used it should be removed in next major version

https://github.com/node-saml/node-saml/blob/0d3d3fae7328ca6a7aaaf36dfe030b8f1baf7da4/src/saml.ts#L514-L623

cjbarth:
As was pointed out in [this](https://github.com/node-saml/passport-saml/issues/909#issuecomment-2048493744) comment, the changes made in #214 were incomplete. This PR addresses this, but doing so is a breaking change.

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -513,11 +513,7 @@ class SAML {
     );
   }
 
-  async getAuthorizeUrlAsync(
-    RelayState: string,
-    host: string | undefined,
-    options: AuthOptions,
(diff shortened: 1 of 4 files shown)


markstos (approved):
## QA Log

- Agree with removing unused args as a breaking change.

PR #384, opened 2025-01-21
docs: Update README.md set never default validateInResponseTo

kdhttps:
# Description

Update Readme.md. Add details in docs that `never` is a default value for validateInResponseTo.

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [ ]
- Tests included? [ ]
- Documentation updated? [x]


markstos (approved):
## QA Log

* Confirmed this is the default: https://github.com/node-saml/node-saml/blob/0d3d3fae7328ca6a7aaaf36dfe030b8f1baf7da4/src/saml.ts#L136

PR #387, opened 2025-02-13
Add support for sha256-rsa-MGF1 signing algorithm

cornzz:
# Description

This PR adds support for the http://www.w3.org/2007/05/xmldsig-more#sha256-rsa-MGF1 signing algorithm.
The xml-crypto library needs to be updated as well, see https://github.com/node-saml/xml-crypto/issues/328 https://github.com/node-saml/xml-crypto/pull/488

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [x]
- Tests included? [ ]
- Documentation updated? [x]

## Summary by CodeRabbit

* **New Features**
  * Added support for the "sha256-mgf1" signature algorithm, enabling use of RSA-PSS signatures with SHA-256 for SAML requests.

* **Documentation**
  * Updated documentation to reflect support for the new "sha256-mgf1" signature algorithm, including configuration examples and detailed explanations.


cjbarth:
Can you add tests to exercise this? It will also make sure it isn't accidentally removed in the future.

kaibernhard:
> Can you add tests to exercise this? It will also make sure it isn't accidentally removed in the future.

Hi @cjbarth! Would a test verifying a signature (created with an external tool) on, let's say `test/static/signatures/valid/response.root-signed.assertion-unsigned.sha256-rsa-mgf1.xml` be sufficient, or do you have anything else in mind?

cjbarth:
@kaibernhard , that should be just fine.

kaibernhard:
@cjbarth I prepared a test [here](https://github.com/node-saml/node-saml/compare/master...digitalservicebund:node-saml:sha256-rsa-mgf1-tests).
I assume we first need a release of `xml-crypto`, then the changes in this PR and the test. @cornzz I could create a PR to add the test to this PR, alternatively we can proceed like we did in `xml-crypto` with a separate PR for the test.

cjbarth:
I'm working at getting a release of xml-crypto out the door. There are some PRs that people are actively replying on with me, so I want to make sure that I get them landed before we cut a release.

cornzz:
> @cornzz I could create a PR to add the test to this PR

@kaibernhard sounds good to me! Thank you!

cornzz:
Hi @cjbarth, do you have an idea when this PR could possibly be merged and released?
Also, do you know if the release including this change will also include [this breaking change](https://github.com/node-saml/node-saml/pull/398)? Incompatibility with node 18 could currently cause problems for us 😅

PR #396, opened 2025-07-15
Add support for AttributeConsumingService in SAML metadata

markstos:
This adds support for rich AttributeConsumingService metadata, fixing [#406](https://github.com/node-saml/passport-saml/issues/406).

While it's "just XML" and folks could create their own, this approaches improves the developer experience with TypeScript and rich JSdoc hints that IDEs can service to developers to make the experience more efficient and enjoyable than crafting raw
XML by hand.

To support that, detailed JSDoc and TypeScript docs are included.

## Recommend Testing

Try specifying the new attribute in an IDE like VS Code and check the experience. Intellisense should provide lots of hints
and docs for this now.

## Context

[SAML 2.0 Metadata Specification](https://docs.oasis-open.org/security/saml/v2.0/saml-metadata-2.0-os.pdf)

Key Sections:
•  Section 2.4.4: "Element <md:AttributeConsumingService>" (pages 23-25)
•  Section *******: "Element <md:ServiceName>" 
•  Section *******: "Element <md:ServiceDescription>"
•  Section *******: "Element <md:RequestedAttribute>"

XML Schema:
The XML Schema definition for AttributeConsumingService can be found in the SAML 2.0 Metadata Schema:
[saml-schema-metadata-2.0.xsd](https://docs.oasis-open.org/security/saml/v2.0/saml-schema-metadata-2.0.xsd)

Key Points from the Spec:
•  AttributeConsumingService is used by service providers to describe the attributes they would like to receive
•  It's optional but highly recommended for proper attribute handling
•  Each service can have multiple AttributeConsumingService elements with different purposes
•  The index attribute is required and must be unique within the SPSSODescriptor
•  The isDefault attribute indicates the default service for attribute consumption

The specification defines how identity providers should interpret these elements when generating attribute statements in SAML assertions.

## Summary by CodeRabbit

* **New Features**
  * Added support for defining multiple SAML attribute-consuming services in service provider metadata.
  * Added standardized SAML attribute names and name formats, with support for custom values.
  * Added service selection, localized names and descriptions, requested attributes, friendly names, and required/optional attribute settings.

* **Bug Fixes**
  * Corrected handling of empty service configurations and explicitly defined default flags.
  * Preserved the required ordering of metadata elements.

* **Documentation**
  * Added configuration guidance and examples for attribute-consuming services.


srd90:
Update: it seems that description has been edited to point to correct issue. Following comment is now outdated:

---

Sidenote: referenced issue (406) at first chapter of description does not exist at `node-saml/node-saml` repository (at the time latest issue/pr/discussion number is https://github.com/node-saml/node-saml/pull/396) and thus it is not rendered as link at the monent.

It (must) refer to `node-saml/passport-saml` issue:
* https://github.com/node-saml/passport-saml/issues/406

I.e. consider editing description so that issue number is replaced with URL to passport-saml repository's https://github.com/node-saml/passport-saml/issues/406.

cjbarth:
@markstos , is what @srd90  says about the issue you're referencing correct?

jetze-meilink:
Would be great to have this feature, we're currently waiting for this.
Let me know if I can help with anything.

src/types.ts
```diff
@@ -125,6 +125,57 @@ export interface XmlSignatureLocation {
...
+  /** Basic format */
+  BASIC: "urn:oasis:names:tc:SAML:2.0:attrname-format:basic",
+  /** Unspecified format */
+  UNSPECIFIED: "urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified",
+} as const;
+
+/**
+ * Common language codes for ServiceName and ServiceDescription
+ */
+export type CommonLanguageCode = "en" | "es" | "fr" | "de" | "it" | "pt" | "ja" | "ko" | "zh" | "ar" | "ru";
```
srd90: Out of curiosity:

From what perspective these are most common language codes? :)

I see that list includes e.g. `it` but not e.g. `hi`, `vi` or `id`

References:
* https://en.wikipedia.org/wiki/List_of_languages_by_total_number_of_speakers
* https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes

---

Real question: what purpose does aforementioned 11 out of 31 languages (by total number of speakers and `it` being 31st) serve?

jetze-meilink:
I tested it out and it works as expected. :ok_hand:

src/types.ts
cjbarth: I'm in agreement here, I've deleted all references to `CommonLanguageCode` (an import, and export, and this definition) and the code still compiles just fine. I can't find any reference to this type anywhere useful. What purpose does having this definition serve @markstos ?

jetze-meilink: Can't we leave this out for now? I don't see it used as both of you mentioned and it seems to be the only blocking factor in getting this merged?

cjbarth: Resolved in 1e0e5ea: `CommonLanguageCode` is removed, along with its import and export.

To give the reasoning rather than just "it was unused" — the problem @srd90 identified is that the subset implies a closure BCP 47 does not have. `xml:lang` accepts any well-formed language tag, so there is no principled place to stop adding entries, and a union invites the question of why `it` is present and `hi` is not. The field stays `string`, and the JSDoc now points at BCP 47 with `en-GB` and `zh-Hant` among the examples so that region and script subtags read as expected rather than looking unsupported.

We kept the same test in the other direction for `@NameFormat`. SAML 2.0 Core, section 8.2 defines a closed set of three name formats, so that one is now typed as a union, and `@Name` offers the common OIDs as completions. Both use `| (string & {})`, so an unlisted value still type-checks. The rule we settled on is that types enumerate what the specification closes, and JSDoc documents what it leaves open.

Thanks for pressing on this one.

cjbarth:
I have pushed a round of changes to this branch and merged `master` in. Tests are at 273 passing with lint and formatting clean. Summary of what changed and why, since some of it removes work that was already here.

## Removals

**`RequestedAttribute.AttributeValue`.** This emitted `<AttributeValue>` into the metadata namespace, but `md:RequestedAttributeType` extends `saml:AttributeType`, so the element belongs in `urn:oasis:names:tc:SAML:2.0:assertion`. The `EntityDescriptor` only declares `xmlns` and `xmlns:ds`, so anything produced through this field was schema-invalid and would be rejected on upload by a validating IdP. Nothing documented or tested it. Fixing it properly means adding an `xmlns:saml` declaration to every generated document, which is more surface than an untested field justifies, so I removed it. @markstos, if you had a use case in mind — Shibboleth value-constrained requested attributes are the usual one — say so and we can add it back correctly; it is additive either way.

**`CommonLanguageCode`.** Removed, per the thread above.

## Kept, and now wired in

`COMMON_SAML_ATTRIBUTES` and `SAML_ATTRIBUTE_NAME_FORMATS` stay, but they were exported without being referenced by any field, so they delivered no completions. `@Name` is now `CommonSamlAttributeName | (string & {})` and `@NameFormat` is `SamlAttributeNameFormat | (string & {})`, which offers the known values while still accepting anything else.

## Fixes

- `@isDefault` is emitted whenever it is defined, so an explicit `false` survives. It previously round-tripped differently from `@isRequired` in the same feature.
- `@NameFormat` was documented as defaulting to the URI format, but nothing substitutes it. Omission emits no attribute, which SAML 2.0 Core, section 2.7.3.1 reads as `unspecified` — the one direction that breaks attribute matching at IdPs that distinguish the two. The documentation now says what happens, and every example requesting an OID-named attribute sets the format explicitly.
- The unrelated change to `getAuthorizeFormAsync` is reverted. Every value in `samlMessage` is already coerced upstream, so the added `|| ""` was unreachable, and it would have turned a legitimate `0` into an empty string. As a side effect `src/saml.ts` leaves this PR entirely, which is why merging `master` — including the `debug` removal that rewrote that file — was conflict-free.
- The `SamlOptions` doc comment had been absorbed by the new interface, leaving the most-used public type undocumented. Restored.

## Tests

Folded `test/attribute-consuming-service.spec.ts` into the existing `generateServiceProviderMetadata` suite and assert against a golden file, matching how `metadataContactPerson` and `metadataOrganization` are covered. This matters beyond consistency: the previous tests parsed with `xml2js`, which discards sibling order, so they could not have caught a regression in element sequencing — and the sequencing is load-bearing, since both `SPSSODescriptorType` and `AttributeConsumingServiceType` are `xs:sequence`. The second service in the fixture deliberately declares its keys out of order so the golden file fails if the ordering logic is ever tidied away.

The test also sets `attributeConsumingServiceIndex` alongside a matching declaration, which is the asymmetry this feature exists to close: we could already reference a service index from an `AuthnRequest` without any way to declare what that index meant.

## Deliberately deferred

No runtime validation of `@index` uniqueness, non-array input, or the `unsignedShort` range. `metadataContactPerson` and `metadataOrganization` validate nothing either, so doing it here alone would be inconsistent; better as one follow-up covering all three. The constraints are documented in the README meanwhile. Happy to be overruled on this.

## Note on node-saml/passport-saml#406

Merging this will not close that issue. The option reaches passport-saml structurally — `PassportSamlConfig = SamlConfig & StrategyOptions` and the strategy delegates straight to `generateServiceProviderMetadata` — so no code change is needed there, but it does need a node-saml release and then a passport-saml bump before @aeaton-overleaf can use it.

PR #397, opened 2025-07-21
Use new .signedReferences interace in xml-crypto to "see what is signed"

ahacker1-securesaml:
I added a new feature on xml-crypto, that returns the exact bytes that were signed. You can ONLY use the exact bytes that were signed by the SignedXML crypto and nothing else.

I then integrated this new feature into the node-saml library with backwards compatibility.

Security fix, apply see what is signed method

@srd90, @markstos can you review the PR.

This fixes the security vulnerability by loading from only signed contents.

To review:

We only care about SAML assertions/ authentication responses. I don't care about logout request/logout responses, because they aren't sensitive.
Security: Whether the SAML assertion FOR ALL PATHS is loaded from only signed contents.
Check whether getVerifiedXML returns only data from getSignedReferences()
Check whether there are code paths to change profile that are loaded from unverified XML.


cjbarth:
When I get back to my normal computer later today I can address those DeepScan issues if you don't get to it before then.

cjbarth:
This has been released as 5.1.0. Since there were some other changes, that was the correct semver, but it should be compatible with the standard `package.json` version identifier: `^`.

When @srd90 gets back in to his normal routine, I'm looking forward to his comments. I'm also looking forward to how we can clean all this up for a v6 release.

PR #399, opened 2025-07-27
Set `inResponseTo` validation to `always`

cjbarth:
# Description

This tightens `inResponseTo` validation to `always` based on [this](https://github.com/node-saml/node-saml/pull/314#issuecomment-1684289377) comment.

## Summary by CodeRabbit

* **Documentation**
  * Updated the documentation to reflect the new default behavior for the `validateInResponseTo` setting.

* **Bug Fixes**
  * Adjusted tests to explicitly disable `InResponseTo` validation where necessary, ensuring consistent test results.

* **Refactor**
  * Changed the default behavior to always validate the `InResponseTo` attribute in SAML responses unless specified otherwise.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,8 @@ const saml = new SAML(options);
 
 - **InResponseTo Validation**
 - `validateInResponseTo`:
-  - if `"always"`, then InResponseTo will be validated from incoming SAML responses
-  - if `"never"`(default), then InResponseTo won't be validated.
+  - if `"always"` (default), then InResponseTo will be validated from incoming SAML responses
+  - if `"never"`, then InResponseTo won't be validated.
   - if `"ifPresent"`, then InResponseTo will only be validated if present in the incoming SAML response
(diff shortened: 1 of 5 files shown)


markstos (approved):
The new default ffollows the philosophy of "secure by default".

ahacker1-securesaml:
Attackers can just exclude inResponseTo (and there's no mechanism to reject SAML responses without InResponseTo). So this doesn't really make it more secure.

cjbarth:
With this default, the setting of `always` will require that `inResponseTo` be set. If we set this parameter to `ifPresent`, then the attacker could leave it off. So, unless there's something else I'm missing @ahacker1-securesaml , this does increase security.

ahacker1-securesaml:
I checked the code, and yes @cjbarth you are correct.

However, many applications use IdP initiated SSO, which has SAML responses without InResponseTo. I agree that this has some security benefit, namely preventing login CSRF.

When you click the Okta/Entra app tiles and the application logs in automatically you are basically using IdP initiated SSO. How many applications you use support this flow? Your change would basically break this flow.

I would instead change it to ifPresent, and then add some documentation saying the risks of IdP initiated SSO/login CSRF, and recommend changing it to always

PR #406, opened 2025-09-08
security: remove debug dependency

ahacker1-securesaml:
# Description

Removes `debug` dependency, see https://www.aikido.dev/blog/npm-debug-and-chalk-packages-compromised

while malware has been removed, we should be reducing dependencies in general to improve security (prevents threat of future compromise) 

## Summary by CodeRabbit

* **New Features**
  * Added a public method to generate service provider metadata for integrations.

* **Refactor**
  * Switched internal debug logging mechanism and reduced noisy logs; preserved runtime behavior.
  * Minor error-message text correction in XML handling.

* **Chores**
  * Removed unused debug-related dependency.
  * Tightened linting to treat console usage as an error.


markstos (commented):
I agree with the rationale here.

And while leaving commented out code usually is a code smell. Here it's helpful if we want to temporarily add back a debug function. 

Another option would be a zero-dependency equivalent, like:

```javascript
if (process.env.DEBUG='true') {
   console.error('Some debug logging');
}
```
@cjbarth Which approach appeals to you?

cjbarth:
@markstos , why wouldn't we just use [`console.debug()`](https://developer.mozilla.org/en-US/docs/Web/API/console/debug_static)? For cases where we aren't serializing a lot of data, the overhead should be minimal. It would be nice to get rid of that dependency (unless we're going to go full-in on it).

This is preferable to me than uncommented code because this is a library and I doubt people will be modifying the code just to do some testing/debugging.

ahacker1-securesaml:
I left it commented out, so maintainers can choose a strategy that is best.

Let's use console.debug:

```js
if (process.env.NODE_SAML_DEBUG='true') {
   console.debug('Some debug logging');
}
```

or
```js
console.debug('...')
```

cjbarth:
There's no need for the `if` guard IMO.

ahacker1-securesaml:
Looks good to me!

cjbarth:
> `225-235`: **Missing XPath-injection guard in validateXmlSignatureWithPemFile.**
> 
> Unlike getVerifiedXml(), this path interpolates refId into an XPath literal without rejecting quotes. A crafted refId containing ' or " could alter the XPath and defeat the intended check. Add the same guard here before building the XPath.
> 
> ```diff
>    const refId = refUri[0] === "#" ? refUri.substring(1) : refUri;
> +  // prevent XPath injection
> +  if (refId.includes("'") || refId.includes('"')) {
> +    throw new Error("ref URI included quote character ' or \". Not a valid ID, and not allowed");
> +  }
>    // If we can't find the reference at the top level, reject
>    const idAttribute = currentNode.getAttribute("ID") ? "ID" : "Id";
> ```

@ahacker1-securesaml  or @srd90, what do you think of this suggestion from coderabbitai?

ahacker1-securesaml:
> validateXmlSignatureWithPemFile

I believe that this was an old method used to verify signatures. Now we have getVerifiedXML(), which has been tested to work. we can remove the old method, and replace usages with getVerifiedXML(). I will do this in a separate PR.

cjbarth:
@coderabbitai review

PR #408, opened 2025-09-11
Feat/useNewVerification

ahacker1-securesaml:
# Description

Thank you for taking the time to create a PR to support this community-maintained project. We are all volunteers here, so features are only added by contributors like you. Please also be patient while the maintainers work to review your proposed changes for quality and efficacy.

Please include a _summary of the change_ and, where appropriate, _which issue is addressed_.

Your description should also include a _comment about the use-case_ for this change, including what systems you may be interfacing with that require this change. This will help future users know how this library is being used in the wild and how well supported it is.

If the change involves the SAML specification, please _include a link to the relevant part(s) of the SAML specification_. Doing so speeds up the review process and helps the maintainers make sure that this project remains spec-compliant. You might start [here](https://www.oasis-open.org/standards#samlv2.0) to find the part of the specification that relates to this PR.

_Please include tests._ Doing so will ensure that the changes made in this PR are not undone or otherwise corrupted by future changes.

# Checklist:

- Issue Addressed: [ ]
- Link to SAML spec: [ ]
- Tests included? [ ]
- Documentation updated? [ ]

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,6 @@
         "@types/xml2js": "^0.4.14",
         "@xmldom/is-dom-node": "^1.0.1",
         "@xmldom/xmldom": "^0.8.10",
-        "debug": "^4.4.0",
         "xml-crypto": "^6.1.2",
         "xml-encryption": "^3.1.0",
         "xml2js": "^0.6.2",
@@ -3100,6 +3099,7 @@
       "version": "4.4.0",
       "resolved": "https://registry.npmjs.org/debug/-/debug-4.4.0.tgz",
       "integrity": "sha512-6WTZ/IxCY/T6BALoZHaE4ctp9xm+Z5kY/pzYaCHRFeyVhojxlrm+46y68HA6hr0TcwEssoxNiDEUJQjfPZ/RYA==",
+      "dev": true,
       "license": "MIT",
       "dependencies": {
         "ms": "^2.1.3"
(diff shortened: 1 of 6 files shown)


cjbarth:
Picking this up for tracking rather than letting it go stale. Thank you for it — the core idea
is right, and it identifies something real.

`validatePostRequestAsync` verifies a signature and then acts on the document parsed from the
unverified input, which is exactly the shape `AGENTS.md` says we are moving away from. I have
filed #428 for that fix and #429 for the `validateSignature` removal this PR also contains.

**Milestoned v6**, because as it stands this removes the `validateSignature` export, and that
half cannot ship in a minor. If it is split, the call-site migration moves to 5.2.0 under #428
and only the export removal stays here — that ordering is what `AGENTS.md` asks for, since the
`@deprecated` tag cannot go on until the last internal caller is gone.

Two notes for whoever takes it forward:

- `dom` is still parsed from the raw XML in this diff, and it is what `getNameIdAsync()` reads.
  The `NameID` is the identity being logged out, so it needs to come from the verified bytes
  too. #428 covers both.
- This predates 25c434a, which removed the `debug` dependency in favour of `util.debuglog`, so
  it no longer applies cleanly. The commented-out `debug()` calls here should become
  `debugLog()` calls rather than disappear.

For context, there is related work in flight for 5.2.0, all of it warning about things the next
major changes: #420 (unsigned redirect-binding logout), #421 (the unused `host` argument), #425
(SHA-1 and replay-protection defaults), #426 (`Profile.getSamlResponseXml()`) and #427
(attributes with no usable value). #428 is the remaining piece and is the one this PR overlaps.

PR #413, opened 2026-06-05
fix: keep empty and valueless SAML attributes in profile.attributes

### LINKED ISSUE #227: [BUG] Empty and Null Attributes are not returned in profile.attributes
**To Reproduce**

Encounter a SAML assertion containing an empty Attribute tag with no nested AttributeValue.

example:`<saml2:Attribute Name="employee_id"/>`

I've encountered these empty tags in the wild using Google Workspaces as a SAML IDP

```typescript
        it("A null value given with an object should be null", async () => {
          const xml =
            '<Response xmlns="urn:oasis:names:tc:SAML:2.0:protocol" ID="response0">' +
            '<saml2:Assertion xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion" Version="2.0">' +
            "<saml2:AttributeStatement>" +
            '<saml2:Attribute Name="attributeName" ' +
            'NameFormat="urn:oasis:names:tc:SAML:2.0:attrname-format:unspecified" />' +
            "</saml2:AttributeStatement>" +
            "</saml2:Assertion>" +
            "</Response>";

          const signingKey = fs.readFileSync(__dirname + "/static/key.pem");
          const signingCert = fs.readFileSync(__dirname + "/static/cert.pem", "utf-8");
          const signedXml = signXmlResponse(xml, { privateKey: signingKey });

          const base64xml = Buffer.from(signedXml).toString("base64");
          const container = { SAMLResponse: base64xml };
          const samlObj = new SAML({
            cert: signingCert,
            audience: false,
            issuer: "onesaml_login",
            wantAssertionsSigned: false,
          });
          const { profile } = await samlObj.validatePostResponseAsync(container);
          assertRequired(profile, "profile must exist");
          expect(profile.attributes?.attributeName).to.be.null;
        });
```

**Expected behavior**

The current behavior of omitting the value entirely makes it difficult to determine if the attribute is configured at all by the IdP, or 
if the data associated with the user is incorrect. For example `{first_name: null}` means that the user doesn't have a first name, while `{}` means that maybe the user doesn't have a first name, maybe the attribute isn't configured properly. 

## No Attribute Value
SAML Core line 1219 states
> Within an `<AttributeStatement>`, if the SAML attribute exists but has no values, then the `<AttributeValue>` element MUST be omitted. 

I believe that the most idiomatic way to represent something that exists but has no values - e.g. `<saml2:Attribute Name="employee_id"/>` - would be as ```{ "employee_id": null }```.

## Empty Attribute Value
SAML Core line 1246 states:
> If a SAML attribute includes an empty value, such as the empty string, the corresponding
`<AttributeValue>` element MUST be empty (generally this is serialized as `<AttributeValue/>`).
This overrides the requirement in Section 1.3.1 that string values in SAML content contain at least one
non-whitespace character.

I believe the test case `An undefined value given with an object should still be undefined` is incorrect - the empty attribute with type `xs:string` should be returned as an empty string literal (`''`), not as `undefined`.

I'm more than happy to update the code + testcases - I understand these would be breaking changes - so please let me know if there is a preferred way to structure the PRs / some future branch I should target. 

**Environment**

- Node.js version: 18
- @node-saml/node-saml version: 4.0.2

arpitjain099:
Fixes #227

## Problem

When a SAML assertion carries an attribute that exists but has no real value, node-saml currently loses that information in `profile.attributes`:

- An attribute with no `AttributeValue` child at all, for example `<saml2:Attribute Name="employee_id"/>`, is dropped from `profile.attributes` entirely. Google Workspace emits these in the wild.
- An attribute whose `AttributeValue` is present but empty, for example `<AttributeValue/>` with `xsi:type="xs:string"`, comes back as `undefined`.

In both cases a consumer cannot tell a valueless attribute apart from one the IdP never sent: `{}` and `{ employee_id: null }` should not collapse to the same thing.

The SAML 2.0 Core spec is explicit about both shapes:

- Section 1219: within an `<AttributeStatement>`, if the SAML attribute exists but has no values, the `<AttributeValue>` element MUST be omitted. The attribute itself still exists.
- Section 1246: if a SAML attribute includes an empty value such as the empty string, the corresponding `<AttributeValue>` element MUST be empty (serialized as `<AttributeValue/>`).

## Fix

`src/saml.ts` attribute extraction now keeps these attributes instead of discarding them:

- No `AttributeValue` child is represented as `null` (the attribute exists, no value).
- An empty `AttributeValue` is represented as the empty string `""` rather than `undefined`, matching the `xs:string` empty-value semantics from section 1246.

Populated single-value, object-valued (nested `NameID`), and multi-valued attributes are parsed exactly as before. The single-value-versus-array branch is untouched.

## Before / after

For `<saml2:Attribute Name="employee_id"/>`:

- before: `profile.attributes` does not contain `employee_id`
- after: `profile.attributes.employee_id === null`

For `<saml2:Attribute Name="x"><AttributeValue/></saml2:Attribute>`:

- before: `profile.attributes.x === undefined`
- after: `profile.attributes.x === ""`

## Tests

Added two tests under the `validatePostResponse xml signature checks` suite, including the exact null-attribute case from the issue:

- "An attribute with no AttributeValue should be null"
- "An empty AttributeValue should be an empty string"

Both fail on `master` (the first throws `expected {} to have property 'attributeName'`, the second `expected undefined to equal ''`) and pass with this change. I also updated the two existing tests that asserted the old drop/undefined behavior, since the issue calls that behavior out as incorrect:

- `accept response with an attributeStatement element without attributeValue` now expects `evilcorp.roles` to be present as `null`.
- the former `An undefined value given with an object should still be undefined` is renamed to `An empty AttributeValue should be an empty string` and asserts `""`.

Full suite is green (273 passing), and lint and prettier pass on the changed files.

## Compatibility note

This is a behavior change for consumers that relied on valueless attributes being absent or `undefined`, so it fits a semver-major release. That lines up with the breaking-change release discussed in the issue thread.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SAML attribute handling to properly preserve missing attributes with `null` values instead of omitting them.
  * Fixed attribute value mapping to correctly return empty strings for empty `AttributeValue` nodes.

* **Tests**
  * Updated test cases to validate new SAML attribute processing behavior.

### DIFF
--- a/src/saml.ts
+++ b/src/saml.ts
@@ -1167,19 +1167,35 @@ class SAML {
         const hasChildren = Object.keys(value).some((cur) => {
           return cur !== "_" && cur !== "$";
         });
-        return hasChildren ? value : value._;
+        if (hasChildren) {
+          return value;
+        }
+        // An attribute that exists but holds an empty value (for example
+        // `<AttributeValue/>` for an `xs:string`) has no character data, so
+        // xml2js leaves `_` undefined. SAML Core section 1246 says such a value
+        // is the empty string, so represent it as "" rather than dropping it.
+        return value._ ?? "";
       };
 
       if (attributes.length > 0) {
         const profileAttributes: Record<string, XMLValue | XMLValue[]> = {};
 
         attributes.forEach((attribute) => {
+          const name: string = attribute.$.Name;
+
+          // An attribute may exist with no `AttributeValue` child at all (for
+          // example `<Attribute Name="x"/>`). SAML Core section 1219 allows
+          // this for an attribute that exists but has no values. Represent it as
+          // null so consumers can tell it apart from a missing attribute,
+          // instead of dropping it from the profile entirely.
           if (!Object.prototype.hasOwnProperty.call(attribute, "AttributeValue")) {
-            // if attributes has no AttributeValue child, continue
+            profileAttributes[name] = null;
+            if (!Object.prototype.hasOwnProperty.call(profile, name)) {
+              profile[name] = null;
+            }
             return;
           }
 
-          const name: string = attribute.$.Name;
           const value: XMLValue | XMLValue[] =
             attribute.AttributeValue.length === 1
               ? attrValueMapper(attribute.AttributeValue[0])
--- a/test/tests.spec.ts
+++ b/test/tests.spec.ts
@@ -839,8 +839,12 @@ describe("node-saml /", function () {
         expect(profile.issuer).to.equal("https://evil-corp.com");
         expect(profile.nameID).to.equal("vincent.vega@evil-corp.com");
         expect(profile).to.have.property("evil-corp.egroupid", "vincent.vega@evil-corp.com");
-        // attributes without attributeValue child should be ignored
-        expect(profile).to.not.have.property("evilcorp.roles");
+        // An attribute that exists but carries no AttributeValue child is kept
+        // with a null value rather than dropped, so callers can distinguish a
+        // valueless attribute from one the IdP never sent.
+        expect(profile).to.have.property("evilcorp.roles", null);
+        const attributes = profile.attributes as Record<string, unknown>;
+        expect(attributes).to.have.property("evilcorp.roles", null);
       });
 
       it("valid xml document with multiple SubjectConfirmation should validate", async () => {
@@ -1509,7 +1513,7 @@ describe("node-saml /", function () {
           });
         });
 
-        it("An undefined value given with an object should still be undefined", async () => {
+        it("An empty AttributeValue should be an empty string", async () => {
           const xml =
             '<Response xmlns="urn:oasis:names:tc:SAML:2.0:protocol" ID="response0">' +
             '<saml2:Assertion xmlns:saml2="urn:oasis:names:tc:SAML:2.0:assertion" Version="2.0">' +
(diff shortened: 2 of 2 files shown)


cjbarth:
Two things to fix before this can land, plus one design question.

**1. `xsi:nil` becomes `""`.** SAML Core §2.7.3.1.1 distinguishes an empty value (`<AttributeValue/>`, the empty string) from a null one (an empty `AttributeValue` carrying `xsi:nil="true"` or `"1"`). `attrValueMapper` ignores `$`, so both have no `_`, and `value._ ?? ""` makes both `""`. A null value should be `null`.

The prefix can't be matched by name. xml2js keeps attribute prefixes as written, and `xsi` is only a convention. #427 adds `isXsiNil(value, [assertion, statement, attribute])` to `src/saml.ts`, which resolves the prefix against the in-scope `xmlns:` declarations. Under inclusive c14n the declaration stays on an ancestor, so reading only the value's own `$` isn't enough. Please reuse that helper once #427 merges.

**2. `<Attribute/>` now throws.** It parses to the string `""`, so `const name: string = attribute.$.Name`, moved above the `AttributeValue` check, throws a `TypeError`. On `master` such an element is skipped. `Name` is required by the schema, but this change shouldn't turn a tolerated document into an exception as a side effect.

**Question: zero values vs. a null value.** With (1) fixed, `<Attribute Name="x"/>` (no values) and `<Attribute Name="x"><AttributeValue xsi:nil="true"/></Attribute>` (one null value) both become `null`. The spec treats these as different states. `[]` for "no values" would keep them apart, but `[]` is truthy, so `if (profile.x)` would pass for an attribute with no values. We'll let you know which way to go.

PR #416, opened 2026-09-16
Docs: add AGENTS.md and CLAUDE.md

cjbarth:
# Description

Adds contributor guidance for AI coding agents, modeled on the files in [xml-crypto](https://github.com/node-saml/xml-crypto) so the two projects read as a pair. `CLAUDE.md` is a pointer; all the guidance lives in `AGENTS.md`.

**Use case:** agents contributing here are, in practice, reading `README.md` and imitating whatever they find nearby. In a library whose job is deciding whether an assertion is trustworthy, imitation is the wrong default — a lot of what is nearby predates the standards we would apply today.

So `AGENTS.md` describes the library we intend to have rather than the one years of mixed contributions produced, and says so explicitly: the file wins over precedent, and code that contradicts it is named as a debt rather than left to be copied. The debts it names are the boolean `validateSignature`, the unsigned XML reachable through `Profile.getSamlResponseXml` (which `src/saml.ts` already comments "should be deprecated"), and the SHA-1 defaults in `initialize` and `algorithms.ts`.

Three positions worth review, since they are opinions and not descriptions:

- **Security-relevant options should have no default.** Absence should be a `TypeError` at construction, as `callbackUrl`, `issuer`, and `idpCert` already are. An integrator who has to name their signature algorithm has thought about it once; one who inherits `sha1` has not thought about it at all. Note this includes `getSigningAlgorithm`/`getDigestAlgorithm` falling through to SHA-1 on an unrecognized value, so a typo currently downgrades the caller silently.
- **A defense invariant no test can fail is not an invariant.** Each rule names where its test lives, and a new rejection needs a fixture under `test/static/signatures/invalid/`.
- **A deprecation process**, since we do have candidates. It records that `@deprecated` tags and a lint rule forbidding deprecated usage have to land together or the build breaks on our own call sites — which is why `validateSignature` carries a comment instead of a tag today — and that deprecating a *default* is harder than deprecating an API, because silence is what is being removed and there is no call site to warn at.

## One behavior change

`test/static` is now excluded from Prettier. Canonicalization and digests depend on the exact fixture bytes, and until now those were protected only by the accident that Prettier ships no XML parser. Making it explicit keeps `prettier --write .` unconditionally safe and stops a future XML plugin from silently invalidating every signature in the suite. `npm run lint` still passes.

# Checklist:

- Issue Addressed: n/a
- Link to SAML spec: n/a — no behavior change
- Tests included? n/a — documentation, plus a Prettier scope change. `npm test` (272 passing) and `npm run lint` both pass.
- Documentation updated? [x] — this PR is the documentation

## Summary by CodeRabbit

- **Documentation**
  - Added comprehensive repository guidance covering supported runtimes, commands, API compatibility, security practices, testing standards, configuration defaults, and contribution conventions.
  - Added a pointer directing tools and contributors to the central guidance document.

- **Chores**
  - Excluded signed test fixtures from automatic formatting, preserving their exact byte content for consistent validation.


markstos (dismissed):
I checked if it's still necessary to have a stub CLAUDE.md and apparently it is, due to Claude itself.

markstos (approved):
Revision looks good.

The revised text better distinguishes “XML formatting is acceptable” from “signed fixtures must never be reformatted.”

markstos:
Well, as of Claude Code 2.1.277, Claude has added support for AGENTS.md, so CLAUDE.md is not needed after that.

PR #417, opened 2026-09-17
Update dependencies and fix the CI matrix

cjbarth:
# Description

Two commits. The first takes every dependency that can move without crossing a major; the second takes majors for development and build-chain packages that support the Node floor. **Every file in `lib/` is byte-identical to what `master` builds today**, so nothing here reaches consumers.

Production dependencies audit clean. Counting development dependencies, known vulnerabilities go from **42 to 3**; all three are reached through mocha, which is pinned below.

## This also fixes a break already on `master`

`npm ci` fails on npm 11 today, on any branch:

```
npm error code EUSAGE
npm error `npm ci` can only install packages when your package.json and
npm error package-lock.json ... are in sync.
npm error Missing: fsevents@2.3.3 from lock file
```

The lock recorded `chokidar`'s optional `fsevents` dependency without a package entry for it. Older npm tolerated that; npm 11 does not. Regenerating the lock repairs it. The lock had to be rebuilt anyway, because it pinned the `typescript-eslint` trio at 8.30.1 and would not move plugin and parser in lockstep.

## CI matrix

`[18, 20, 22]` plus a `latest` entry becomes `[18, 20, 22, 24, 26]` on both jobs, and `fail-fast: false` is set.

The `latest` entry was labelled `experimental: true`, but nothing consumed the flag -- there is no `continue-on-error` -- so when it broke on a moving target it cancelled the three supported-version jobs alongside it, turning one failure into four red jobs. Explicit versions say what we mean. Actions are updated too: checkout and setup-node to v7, codecov-action to v7, and codeql-action off the deprecated v2 to v4.

## Held back, in three groups

**Consumer-facing, so they want their own review.** `@xmldom/xmldom` and `xml-encryption` are runtime dependencies. Note `@xmldom/xmldom` 0.8 to 0.9 reads as minor, but for a `0.x` package the minor field is the breaking position, and this one is the XML parser on the signature verification path.

**TypeScript** is build tooling but reaches consumers all the same. TypeScript 6 defaults `esModuleInterop` to true, rewriting every namespace import in the emitted JavaScript from `require("crypto")` to `__importStar(require("crypto"))`. That is the entire difference: with the flag pinned off, TypeScript 6 emits byte-identical output to 5.9. But pinning it off is a dead end, since the option is deprecated in 6 and stops functioning in 7. Worth its own pull request, where the emit change is the subject rather than a side effect.

**Pinned by the Node 18 floor, not by preference: `mocha`, `chai`, `eslint`, `eslint-plugin-mocha`.** Mocha 12 `require()`s its own ES modules, and chai 5 and later are ESM only, so both need `require(esm)`: Node 24 and Node 22.12 respectively. On Node 18 mocha 12 throws `ERR_REQUIRE_ESM`; chai 6 fails more obscurely, because the failed `require` makes mocha retry the spec through the ESM loader, which then reports `ERR_UNKNOWN_FILE_EXTENSION` for `.ts`. ESLint 10 and eslint-plugin-mocha 12 declare `^20.19.0 || ^22.13.0 || >=24` and `>=22.0.0` respectively. They do run on Node 18 today, but only because npm treats `engines` as advisory, and a patch release could end that with no semver signal -- so holding them keeps the whole development workflow usable on the floor named in `engines`, the same reason mocha and chai are held. The three remaining audit findings are mocha 11's and clear when mocha can move. **Dropping end-of-life Node versions unpins all four together.**

**Pinned to the floor deliberately: `@types/node`** stays on the 18 line because it defines the API surface `src` compiles against. Raising it would let code that only exists on newer Node compile clean and then fail at runtime on the floor named in `engines`. xml-crypto pins the same way, `@types/node@^16` against `engines >=16`.

## Test changes, all forced by the upgrades

- **sinon 22** bundles fake-timers 15, which fakes `queueMicrotask` by default and so stalls every `await`. The forty call sites that only want a fixed clock now pass `toFake: ["Date"]`; the one in `cache.spec.ts` that genuinely advances timers is untouched.
- **`@xmldom/xmldom` 0.8.10 to 0.8.15**, a patch move inside the existing range that would land on any install, rephrased a parse error that one test had pinned character for character. What the library owes its callers is that malformed XML is rejected and that the error carries the position; the phrasing belongs to xmldom, so the assertion now matches on structure and position instead.
- Three `assert(...)` call sites become `assert.ok(...)`, the alias they were always shorthand for.

# Checklist:

- Issue Addressed: n/a
- Link to SAML spec: n/a -- no behavior change
- Tests included? n/a -- dependency and CI work. Verified on Node 18, 20, 22, 24 and 26: `npm ci`, 272 passing, and lint clean on every one.
- Documentation updated? n/a

## Summary by CodeRabbit

* **Chores**
  * Updated development and runtime dependencies to newer supported versions.
  * Refreshed automated security analysis and build workflow tooling.
  * Expanded compatibility checks across Node.js 18, 20, 22, 24, and 26.

* **Tests**
  * Improved test reliability by updating timer simulation and assertion patterns.
  * Made XML parser error validation resilient to wording changes while preserving location checks.
  * Expanded TypeScript test-environment type support.


markstos (approved):
## QA Log

- Reviewed diff. Looks like  lots of standard maintenance here with most of the "diff" being in huge lock file churn.

PR #427, opened 2026-09-17
Warn when an attribute reaches the profile with no usable value

cjbarth:
# Description

#413 changes how two shapes of `Attribute` reach the profile. Both are silent today, so this
is the notice in front of that change.

```xml
<Attribute Name="roles"/>                              <!-- dropped entirely -->
<Attribute Name="team"><AttributeValue/></Attribute>   <!-- present, but `undefined` -->
```

The first is left out of the profile, so a consumer cannot tell it apart from an attribute the
identity provider never sent. The second arrives as `undefined`, which for most consumers is
the same as absent. #413 makes them `null` and `""` respectively, so both stay distinguishable
from an attribute that was not sent.

Neither is visible to a caller today, and that is exactly the problem: someone reading
`profile.roles` has no way to learn their IdP is sending it. The warnings name the attribute,
so the notice says *which* one rather than only that it happened.

## Use case

An operator whose IdP sends a valueless attribute currently sees nothing at all, and will see
the profile change shape under them at the major. Running with `NODE_DEBUG=node-saml` now tells
them which attribute is affected, in their own traffic, before that happens.

## Notes for the reviewer

**The second shape had no fixture, and it could not simply be added to the existing one.**
`test/static/response-with-uncomplete-attribute.xml` covers `<Attribute Name="evilcorp.roles"/>`
and is signed at the `Response` level, so editing it invalidates the signature. The test
therefore builds its own document: strip the signature, insert the attribute, and re-sign with
the repository's own `key.pem` via `signSamlPost`. That keeps a second signed blob out of
`test/static`, where `.prettierignore` exists precisely because those bytes must not be
touched.

The non-warning test is the one worth reading. It pins the current profile shape — `roles`
absent, `empty` present but `undefined` — so #413 cannot land without a test failing and being
rewritten deliberately, which is what `AGENTS.md` asks of behavior that is scheduled to change.

The empty-value check handles the multi-value case, since `AttributeValue` may repeat and only
one of them may be empty.

This is the last of the four v6 items that had no 5.2 notice. #421 covers the `host` removal,
#420 the unsigned redirect logout, #425 the three security-relevant defaults, #426 the unsigned
response accessor.

# Checklist:

- Issue Addressed: prepares #413
- Link to SAML spec: SAML Core allows an `Attribute` with no `AttributeValue`, and treats an empty one as the empty string — which is what #413 implements
- Tests included? Yes — the current profile shape for both, and a warning for each
- Documentation updated? Yes — new "Attributes with no value" section in `README.md`


cjbarth:
Pushed 117d7f6 with `master` merged in. Changes since the first revision, after an outside review:

1. **`xsi:nil`** gets its own warning ("represents it as null"), and the empty-string one no longer fires for it. The prefix is resolved against the in-scope `xmlns:` declarations, not matched by name. The test covers `<AttributeValue xsi:nil="true"/>` with the declaration on an ancestor (inclusive c14n keeps it there), plus a sibling where `xsi` is bound to another namespace. #413 still turns nil into `""`; raised there.
2. **`<Attribute/>`** parses to `""`, so the first revision's `attribute.$.Name` threw where `master` skips the element. The warning now fires only when there is a `Name`. The profile-shape test includes a nameless `Attribute` and failed with that `TypeError` before the fix.
3. **Wording:** the README no longer says callers cannot act on `undefined`. It says the value is present but dropped by `JSON.stringify` and commonly treated as absent, and keeps "looks the same as not sent" only for the missing-`AttributeValue` case.
4. **Tests:** the response is now built inline and signed with `signXmlResponse`, like the neighbouring empty-`AttributeValue` test, rather than by stripping and re-signing a fixture.

The warnings and README no longer link to #413.

